PR #2, opened 2020-05-22
Code changes for when LArSoft adopts thread-safe clocks/properties services

knoepfel:
This pull request is the equivalent of a feature branch for whenever LArSoft migrates to using the thread-safe `DetectorClocks` and `DetectorProperties` providers and services.

It **should not** be merged into develop until LArSoft migrates.


ascarff:
There's a typo in the product_deps, the larsoft version should be `v09_00_00_rc1`. But with this change sbndcode compiles with no problems against this larsoft version. (Sorry for the delay in testing!)

knoepfel:
Thanks @ascarff for looking into this.

PR #6, opened 2020-06-23
Geometry version v1.5

mrguzzo:
- Update to geometry version v1.5 (`Geometry/gdml/sbnd_v01_05.gdml`)
- Changes in the PDS in the geometry:
  - PMT filled with vacuum
  - X-ARAPUCAs now have a single readout channel --> it changed the total number of optical channels
- Updated channel mapping (`OpDetSim/sbnd_pds_mapping.json`)
- Validation that xarapuca_vuv have about twice as many ADC as arapuca_VUV
- Update simulation to use geometry version v1.5 (`Geometry/geometry_sbnd.fcl`)

I got help from @ikatza


ikatza:
I ran the batch of tests `test_runner develop_test_sbndcode` it failed the two detector simulation ones. Reason being that those tests use largeant files that have the old number of optical channels.

Again this breaks the compatibility of old largeant files.

ikatza:
Running `mrb t` all tests passed:

```
Test project /sbnd/data/users/icaza/work/sbndcode/geometry/tests/b/build_slf7.x86_64
    Start  1: geometry_sbnd
1/8 Test  #1: geometry_sbnd ............................   Passed   18.55 sec
    Start  2: geometry_sbnd_test
2/8 Test  #2: geometry_sbnd_test .......................   Passed   16.30 sec
    Start  3: geometry_iterator_sbnd_test
3/8 Test  #3: geometry_iterator_sbnd_test ..............   Passed    2.98 sec
    Start  4: geometry_iterator_loop_sbnd_test
4/8 Test  #4: geometry_iterator_loop_sbnd_test .........   Passed    3.72 sec
    Start  5: dump_channel_map_sbnd_test
5/8 Test  #5: dump_channel_map_sbnd_test ...............   Passed    3.59 sec
    Start  9: prodsingle_reconstruction_test.sh
6/8 Test  #9: prodsingle_reconstruction_test.sh ........   Passed  146.85 sec
    Start 10: fast_simulation_reconstruction_test.sh
7/8 Test #10: fast_simulation_reconstruction_test.sh ...   Passed  124.15 sec
    Start 11: prodcosmics_cry_test
8/8 Test #11: prodcosmics_cry_test .....................   Passed    8.12 sec

100% tests passed, 0 tests failed out of 8

Label Time Summary:
DEFAULT    = 324.27 sec*proc (8 tests)
RELEASE    = 324.27 sec*proc (8 tests)

Total Test time (real) = 324.94 sec

```

ggamezdiego:
We really need to move to the new geometry, with the correct number of optical channels (a 40% less!). But I would say that before the pull request some validation tests must be shown in the PAT meeting proving, for example, that the updated digitized signals make sense

ikatza:
I'm adding Laura to the reviewers. 

@ggamezdiego or maybe you @lpaulucc could you expand on types of tests and signals would you like to see?

ggamezdiego:
I think that a comparison of the digitised signals after and before the update would validate the changes. We should confirm we get the same results as before, given that the updates shouldn't have modified the output signals, if I am understanding well.

ikatza:
@mrguzzo asked my advise for this, however there are some aspects I'm not sure about. @ggamezdiego would you mind giving some more details about the comparison you suggest?

The test goes something along the lines:
- Generate some events, any type? is there are preferred type of interaction?
- Run largeant, that stage has to be run with each version. Unfortunately, the existing photon simulation is not reproducible because is not seeded properly. Point being: the output of largeant will be somewhat different each time you run it. 
- Detector simulation comes next, again run with each version. The input will already be somewhat different. Output of each version will too

A qualitative comparison can be made, inspecting if the waveforms look sort of the same, is that enough?

ggamezdiego:
I think that crossing muons (mips) "parallel" to the photocathode plane, at different drift distances, could be an easy test-sample (with minimal fluctuation) for the purpose of a comparison. If the muons are high momentum (crossing muons), even with different seeds, the light footprint between the two simulations should be very similar. I also agree that testing at the larg4 and reconstruction level would be optimal. The main variable to pay attention would be the integral of the signals. I think that in a test like this, we need to be sure that all the optical channels are alive, so a channel-based comparison is what we would need.

ikatza:
Marina and I came across an issue with the ARAPUCAS, the problem is that they are being given the sizes of XARAPUCAS.

We've tested our fix with this branch.

I have submitted the fix to larsim now: https://github.com/LArSoft/larsim/pull/34

ggamezdiego (approved):

ikatza:
Let's wait a bit to merge, the fix in larsim should come soon.

ikatza (approved):
Fix got merged to larsim. Approving this one.

PR #16, opened 2020-07-30
Feature/etyley pandora shower

etyley:
Changes to fcl configurations for when TRACS is move to PandoraModularShowerCreation, see pull requests in [larreco](https://github.com/LArSoft/larreco/pull/16) and [larpandora.](https://github.com/LArSoft/larpandora/pull/10)

- PandoraShower now uses the new, modular pandora shower (formally TRACS with some changes to dEdx and added length/opening angle)
- PandoraShowerLegacy uses the old pandora shower with no calorimetry information


absolution1:
As mentioned in the larpandora PR (https://github.com/LArSoft/larpandora/pull/10) I've manually triggered the CI system to include the various larsoft and experiment PR features.
It looks like the only SBND CI warnings are related to data product label names (`pandoraShower` vs `pandoraShowerLegacy`).
So, I think we're all good.

Once the LArSoft PRs get merged into their respective develops then this can go ahead as well.

PR #19, opened 2020-08-07
Use CLHEP::RandGeneral() to produce reproducible rands instead of ROOT machinery

ikatza:
Note that the files which contain the test bench data for the time distributions will need to be updated on sbnd_data.

Old files are not compatible with this change and old code won't be able to run with new files.

This addresses the issue of variations in the CI system, hopefully this is the only thing to change to fix this stage.


absolution1:
Hi @ikatza thanks for addressing these so promptly.
I will take a look later on (I have to run out this morning), but I do already have general Qs/requests.

What about the changes makes old files not compatible with the code and new files not compatible with old code?

Would you mind running a quick reproducibility test.  I think a few `std::cout`s should suffice.  Could you add a `cout` after each random throw to print out the random values.  Then, for the tests:

1) Run the code twice and confirm that you get different cout results each time.
2) Override the detsim fcl with the following line and repeat the two runs, hopefully you get consistent results.
```services.NuRandomService.policy: "perEvent"```

File: sbndcode/OpDetSim/DigiArapucaSBNDAlg.cc [hunk 1]
```diff
@@ -32,9 +32,25 @@ namespace opdet {
     cet::search_path sp("FW_SEARCH_PATH");
     sp.find_file(fParams.ArapucaDataFile, fname);
     TFile* file = TFile::Open(fname.c_str());
-    file->GetObject("TimeArapucaVUV", TimeArapucaVUV);
-    file->GetObject("TimeArapucaVIS", TimeArapucaVIS);
-    file->GetObject("TimeXArapucaVUV", TimeXArapucaVUV);
+
+    const size_t vuv_bins = 150;
```
absolution1, line 36: Is this size guaranteed to always be the same or can updates of the template in `sbnd_data` change the vector size?
Either way, it might be a little bit cleaner and easier to maintain if all array lengths are declared in an anonymous namespace at the top of the implementation file, and declare them as `constexpr` i.e.

```
namespace
{
  constexpr size_t vuv_bins = 150;
  //other constrexprs
} //namespace

File: sbndcode/OpDetSim/DigiArapucaSBNDAlg.cc [hunk 2]
```diff
@@ -32,9 +32,25 @@ namespace opdet {
...
+    std::array<double, vuv_bins>* TimeArapucaVUV_arr;
+    file->GetObject("TimeArapucaVUV", TimeArapucaVUV_arr);
+    TimeArapucaVUV= new CLHEP::RandGeneral(fEngine,
```
absolution1, line 39: Does this need to be a pointer?  Would a reference or a plain ol' data member work here?

sbndcode/OpDetSim/DigiArapucaSBNDAlg.cc
```diff
@@ -32,9 +32,25 @@ namespace opdet {
...
+                                           *reinterpret_cast<double (*)[vuv_bins]>(TimeArapucaVUV_arr->data()),
```
absolution1, line 40: Do you need the `reinterpret_cast`.  I think `std::array::data` would return a `double *` in this case?

sbndcode/OpDetSim/DigiArapucaSBNDAlg.cc
```diff
@@ -32,9 +32,25 @@ namespace opdet {
...
+    file->GetObject("TimeArapucaVIS", TimeArapucaVIS_arr);
+    TimeArapucaVIS= new CLHEP::RandGeneral(fEngine,
+                                           *reinterpret_cast<double (*)[vis_bins]>(TimeArapucaVIS_arr->data()),
+                                           vis_bins);
+    const size_t xvuv_bins = 20;
+    std::array<double, xvuv_bins>* TimeXArapucaVUV_arr;
+    file->GetObject("TimeXArapucaVUV", TimeXArapucaVUV_arr);
+    TimeXArapucaVUV= new CLHEP::RandGeneral(fEngine,
+                                            *reinterpret_cast<double (*)[xvuv_bins]>(TimeXArapucaVUV_arr->data()),
+                                            xvuv_bins);
```
absolution1, line 53: Same comments as above.

sbndcode/OpDetSim/DigiArapucaSBNDAlg.cc
absolution1: More importantly, does the `CLHEP::RandGeneral` RNG provide enough granularity for what you need?  
My understanding is that it linearly interpolates between bins when randomly throwing.  Is that sufficient for this use case?

sbndcode/OpDetSim/DigiArapucaSBNDAlg.hh
```diff
@@ -16,6 +16,7 @@
 #include "CLHEP/Random/JamesRandom.h"
 #include "CLHEP/Random/RandFlat.h"
 #include "CLHEP/Random/RandGaussQ.h"
+#include "CLHEP/Random/RandGeneral.h"
 #include "CLHEP/Random/RandPoissonQ.h"
 #include "CLHEP/Random/RandExponential.h"
```
absolution1, line 21: Are all of these headers still needed?
I haven't looked at the entire header/imp in this much detail but always good to ask about required headers ;)

sbndcode/OpDetSim/DigiArapucaSBNDAlg.hh
```diff
@@ -109,9 +107,9 @@ namespace opdet {
 
     CLHEP::HepRandomEngine* fEngine; //!< Reference to art-managed random-number engine
 
-    TH1D* TimeArapucaVUV; //histogram for getting the photon time distribution inside the Arapuca VUV box (considering the optical window)
-    TH1D* TimeArapucaVIS; //histogram for getting the photon time distribution inside the Arapuca VIS box (considering the optical window)
-    TH1D* TimeXArapucaVUV; //histogram for getting the photon time distribution inside the XArapuca VUV box (considering the optical window)
+    CLHEP::RandGeneral* TimeArapucaVUV; //histogram for getting the photon time distribution inside the Arapuca VUV box (considering the optical window)
+    CLHEP::RandGeneral* TimeArapucaVIS; //histogram for getting the photon time distribution inside the Arapuca VIS box (considering the optical window)
+    CLHEP::RandGeneral* TimeXArapucaVUV; //histogram for getting the photon time distribution inside the XArapuca VUV box (considering the optical window)
```
absolution1, line 112: Referring to my earlier Q, do these need to be pointers?

Looks like almost all other data members start with an `f` as well.

sbndcode/OpDetSim/DigiPMTSBNDAlg.cc
```diff
@@ -38,7 +34,14 @@ namespace opdet {
...
     TFile* file = TFile::Open(fname.c_str());
-    file->GetObject("timeTPB", timeTPB);
+
+    // TPB emission time histogram for pmt_coated histogram
+    const size_t timeTPBBins = 1000;
+    std::array<double, timeTPBBins>* timeTPB_arr;
+    file->GetObject("timeTPB", timeTPB_arr);
+    timeTPB = new CLHEP::RandGeneral(fEngine,
+                                     *reinterpret_cast<double (*)[timeTPBBins]>(timeTPB_arr->data()),
+                                     timeTPBBins);
```
absolution1, line 44: Same comments as for the Arapuca alg implementation file

sbndcode/OpDetSim/DigiPMTSBNDAlg.hh
```diff
@@ -16,6 +16,7 @@
 #include "CLHEP/Random/JamesRandom.h"
 #include "CLHEP/Random/RandFlat.h"
 #include "CLHEP/Random/RandGaussQ.h"
+#include "CLHEP/Random/RandGeneral.h"
 #include "CLHEP/Random/RandPoissonQ.h"
 #include "CLHEP/Random/RandExponential.h"
```
absolution1, line 21: Need em all?

sbndcode/OpDetSim/DigiPMTSBNDAlg.hh
```diff
@@ -117,7 +115,7 @@ namespace opdet {
 
     std::vector<double> wsp; //single photon pulse vector
     int pulsesize; //size of 1PE waveform
-    TH1D* timeTPB; //histogram for getting the TPB emission time for coated PMTs
+    CLHEP::RandGeneral* timeTPB; //histogram for getting the TPB emission time for coated PMTs
```
absolution1, line 118: Needs to be a pointer?

sbndcode/OpDetSim/DigiArapucaSBNDAlg.cc
ikatza: See the new commit. No longer using arrays or the bin size variable.

sbndcode/OpDetSim/DigiArapucaSBNDAlg.cc
ikatza: I did try with other types without much success, pointer worked and it seems an acceptable solution to me.

ikatza: Wether has enough granularity or not I guess right now is a moot point, the data files will need to be updated with in situ measurements and with enough granularity.

sbndcode/OpDetSim/DigiArapucaSBNDAlg.cc
ikatza: Yeah, you're right. Updated the code now.

sbndcode/OpDetSim/DigiArapucaSBNDAlg.cc
ikatza: Idem.

sbndcode/OpDetSim/DigiArapucaSBNDAlg.hh
ikatza: Should I rename these to `fTimeArapucaVUV`, `TimeArapucaVIS`, `fTimeXArapucaVUV` then?

absolution1: Depends on how important consistency is to you.
I would opt for consistency but that is, of course, up to you.

sbndcode/OpDetSim/DigiPMTSBNDAlg.hh
ikatza: Removed the extra on 562a419

sbndcode/OpDetSim/DigiArapucaSBNDAlg.hh
ikatza: Renamed these and others on d708189

sbndcode/OpDetSim/DigiArapucaSBNDAlg.hh
ikatza: Removed the extra on 562a419

sbndcode/OpDetSim/DigiPMTSBNDAlg.hh
ikatza: As far as I can work out yes, they need to be pointers.

sbndcode/OpDetSim/DigiArapucaSBNDAlg.cc
absolution1: Hi @ikatza sorry for the slow response.  In regards to RandGeneral, mu2e provide a nice example of what the random sampling looks like: https://mu2e.fnal.gov/public/hep/computing_retired/RandGeneral.gif
If this seems appropriate then, great!

It sounds like the structure wasn't playing ball with swapping from pointers to objects.  A minor change you could make for 'safety' is to promote those pointers to `std::unique_ptr`s to handle the memory management.  This is completely optional and could be kicked down the road to a further PR

absolution1:
Hi @ikatza I think one final check that could be made is to trigger the CI with this branch of sbndcode, to confirm that there are no surprises.  Would you mind doing that?

sbndcode/OpDetSim/DigiArapucaSBNDAlg.cc
ikatza: I have now changed the pointer to `std::unique_ptr` as suggested in d323c6dd

ikatza:
Doing the cost test I get the desired outcome.

For test 1:
Screenshot 2020-08-14 at 2 26 12 PM

For test 2:
Screenshot 2020-08-14 at 2 26 48 PM

ikatza:
> Hi @ikatza
> Thanks again for doing this.
> 
> I've got a few inline comments but overall I think this looks good (99% of my comments are c++ related rather than physics/RNG related...).
> 
> Following the paper trail backwards, I can see that there are N `HepJamesRandom` engines made for N threads:
> 
> https://github.com/SBNSoftware/sbndcode/blob/2e0c5ea0d54c7c0537555ad75ef5be44d65bad6d/sbndcode/OpDetSim/opDetDigitizerSBND_module.cc#L226
> 
> I see that the engines (and all subsequent RNGs seeded using the engine) are declared as bare pointers on the heap without the complementary `delete` statements. I'm guessing the deletes are deliberately missing as the random service owns the pointers to the engines. If so, you might be able to avoid using `new` by storing the engines (and RNG-throwing classes) as references rather than pointers.

I believe the best for now is to open a separate issue about this. I suspect the best solution will be to use `art` supported multi-threading paradigm with Intel TBB.

I had some success implementing it inside the functions that make the waveform noise, successfully running on multiple threads. However they didn't provide any performance improvements: they were too lightweight to offset the overhead cost of spawning threads.

To transition to the `art` supported paradigm we need all the required services to be thread safe. Can't comment on how long will that take or if they are ready now.

ikatza:
Before the objects that had the distributions were `TH1D`, now they are `std::vector<double>`. 
> 
> What about the changes makes old files not compatible with the code and new files not compatible with old code?
>

ikatza:
Running:
`test_runner develop_test_sbndcode` I got a few warnings, not surprisingly. 

Overall I believe there's nothing crazy wrong and these variations are within the observed variance.

```
Test Suite: develop_test_sbndcode
Expanded: single_gen_quick_test_sbndcode single_g4_quick_test_sbndcode single_detsim_quick_test_sbndcode single_reco_basic_quick_test_sbndcode single_anatree_quick_test_sbndcode nucosmics_gen_quick_test_sbndcode nucosmics_g4_quick_test_sbndcode nucosmics_detsim_quick_test_sbndcode nucosmics_reco_basic_quick_test_sbndcode nucosmics_anatree_quick_test_sbndcode compilation_test_sbndcode

 1/11 Test  #1: single_gen_quick_test_sbndcode ...  Passed    4.56 sec;   20.98 kVs; 170544.00 kb;
 2/11 Test  #2: single_anatree_quick_test_sbndcode ..  Passed    3.79 sec;   17.43 kVs; 226968.00 kb;
 3/11 Test  #3: single_reco_basic_quick_test_sbndcode ..  Passed   45.15 sec;  207.69 kVs; 789176.00 kb;
 4/11 Test  #4: single_detsim_quick_test_sbndcode ..  Warning   50.25 sec;  231.15 kVs; 675772.00 kb;
 5/11 Test  #5: single_g4_quick_test_sbndcode ....  Warning   65.74 sec;  302.40 kVs; 231312.00 kb;
 6/11 Test  #6: nucosmics_gen_quick_test_sbndcode ..  Warning   64.12 sec;  294.95 kVs; 577764.00 kb;
 7/11 Test  #7: nucosmics_anatree_quick_test_sbndcode ..  Passed   20.59 sec;   94.71 kVs; 946016.00 kb;
 8/11 Test  #8: compilation_test_sbndcode ........  Passed   22.00 sec;  101.20 kVs; 519228.00 kb;
 9/11 Test  #9: nucosmics_detsim_quick_test_sbndcode ..  Warning   90.22 sec;  415.01 kVs; 2177992.00 kb;
10/11 Test #10: nucosmics_reco_basic_quick_test_sbndcode ..  Warning  115.89 sec;  533.09 kVs; 2207736.00 kb;
11/11 Test #11: nucosmics_g4_quick_test_sbndcode ..  Warning  564.06 sec; 2594.67 kVs; 1798040.00 kb;

5 tests passed (45%), 0 tests failed, 6 tests with warnings, 0 tests skipped, out of 11
Not updating any reference files
```

The `stdout.log` of the detsims with warnings

`single_detsim_quick_test_sbndcode`
```
18c18
< DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 200
---
> DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 196
41c41
< DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 208
---
> DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 195
```

`nucosmics_detsim_quick_test_sbndcode`
```
24c24
< DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 2216
---
> DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 2237
53c53
< DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 2001
---
> DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 2000
```

These two have warning due to scaled cpu usage:
```
nucosmics_gen_quick_test_sbndcode~101~W~Scaled cpu usage 294.95135880000004 not in 400.0:1500.0 range~Check CPU usage
nucosmics_reco_basic_quick_test_sbndcode~101~W~Scaled cpu usage 533.0928411 not in 550.0:1703.0 range~Check CPU usage
```

The two g4 tests have warnings due to a similar seeding issue that needs tackling in that part of the code.

absolution1:
Hi @ikatza 
Thanks so much.  As briefly discussed on slack, if the same tests could be run one more time to confirm that you see the same differences then I think this is good to go.  I believe that the CI will trigger when the push to develop happens, so there will be an additional test :)

ikatza:
Rerruning only the detsim tests, I get the same sizes of data products as before.

`single_detsim_quick_test_sbndcode`
```
18c18
< DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 200
---
> DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 196
41c41
< DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 208
---
> DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 195
```


`nucosmics_detsim_quick_test_sbndcode`
```
24c24
< DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 2216
---
> DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 2237
53c53
< DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 2001
---
> DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 2000
```

PR #25, opened 2020-09-04
Fix DetectorClocksData in Optical Simulation Code

### LINKED ISSUE #24: DetectorClocksData Broken in Optical Simulation Code
With LArSoft v9, changes to `sbndcode` have been made to comply with the new Services structure. The `DetectorClocksData`, used in the optical simulation code to get the optical clock, is set in the constructor of the LArSoft plugin `opDetDigitizerSBND`, and then passed to the digitizer worker via a reference wrapper `std::cref`. Such reference gets broken as the `DetectorClocksData` only exists in the constructor.

The result is that the current optical simulation code is broken and doesn't produce any waveform, as the optical clock returns `inf` when asked for the clock frequency.

marcodeltutto:
Fixes #24.

This PR removes the reference wrapper around detector clocks data when passed to `opDetDigitizerWorkerThread`, so its content is usable also outside of the constructor.


ikatza:
Hi Marco, do you know if this issue is related to the multi-threaded nature of the worker threads?

Maybe @knoepfel would be interested to be aware of this as he's been making these changes to use the new service is other repositories.

knoepfel:
@ikatza, thanks for the heads up, and my apologies for introducing this bug.   @marcodeltutto has made the correct bug fix in this PR.

marcodeltutto:
Thanks @ikatza and @knoepfel, yes this should not affect the threaded part of the optical simulation.

Issue #28, opened 2020-09-07
Single PE Test Bench Waveform Vector


marcodeltutto:
The single PE waveform vector from the test bench measurement is stored in `sbnd_data` in `OpDetSim/digi_pmt_sbnd.root` with name `SinglePEVec`.

In https://github.com/SBNSoftware/sbndcode/blob/develop/sbndcode/OpDetSim/DigiPMTSBNDAlg.cc#L48 is retrieved with name `SinglePEWave`, which causes a segmentation fault.

ikatza:
Ups yeah my fault, forgot to rename a variable. PR #29 fixes this.

PR #29, opened 2020-09-08
Rename vector object to match that from the file on sbnd_data

ikatza:
To address issue #28.


ikatza:
Can you please review it @marcodeltutto? If it works ok I believe it can be safely merged.

marcodeltutto (approved):
Thank you Iker!

marcodeltutto:
Fixes #28

PR #35, opened 2020-10-05
Add SCE Correction Modules into Base Reco fcl-file and Add New Config Running SCE Corrections

gputnam:


gputnam:
@etyley @absolution1 this adds in the configs for space charge correction reco. Ok to merge in?

absolution1:
Hi @gputnam sorry for the delayed response.
The FCL changes/additions look OK to me, thanks for doing them!
Would you be able to rename and move the `standard_reco2_sbnd_sce.fcl` into the `reco` subdirectory and give it an appropriate name.  I'd prefer if we didn't have multiple 'standard' FCLs to avoid confusion.

gputnam:
Makes sense! Done

absolution1:
Thank you Gray.  One other small thing I just noticed, which might be a bit annoying, is that the data product labels (e.g. pandoraSCECalo`) have mixed upper and lower case whereas traditionally I think all product labels are lower case.  Could you make them lower case?  Then I am happy to approve

gputnam:
Do you mean pandoraSCECalo -> pandorasceCalo or pandorascecalo? 

I think stuff like tracks are "pandoraTrack", not "pandoratrack"

absolution1:
So they are!  The pandora labels are actually at odds with the rest of the labelling....
We should probably sort that out, but not now.  The SCE data product labels are all good.

absolution1 (approved):
Everything looks good.  Thanks for addressing the in-conversation comments.

PR #40, opened 2020-10-09
Close files after using them, explicitly open read only

ikatza:


ikatza:
Would you mind reviewing this @ascarff. There is no change in the outcome of code, it's only to better manage resources.

ascarff (approved):
Looks fine to me!

PR #43, opened 2020-10-11
Solving bug at digitization module (issue 34)

fjnicolas:
Solving issue 34: opDetDigitizerWorker only creates a waveform if there are REFLECTED photons arriving at the PMTs.
Details about the issue and this commit can be found in docdb 19579-v1.

List of changes:

OpDetSim/opDetDigitizerWorker.cc
--Reflected and direct photons for the coated PMTs are stored in the main loop over the photon handles
--PMT IDs which see at least one photon (regardless of the light component) are also stored in the loop over the photon handles (if a PMT receive light from both components, only store the ID once)
--Now the waveforms for the coated PMTs are created out of the main loop

OpDetSim/DigiPMTAlg.{cc, hh}
-- New functions to construct and create the waveforms for the coated PMTs (from the previously-stored photon maps) are added ( ConstructWaveform(Lite)CoatedPMT, CreateWaveform(Lite)CoatedPMT)
-- TPB time currently taking into account at the digitization level, hence added for both visible and vuv components. TPB time set to zero in Geant4 because it is unable to simulate the measured-TPB time spectrum (Phys. Rev. C, 91,035503). This delay (i.e. the measured 4-exponential).

LArSoftConfigurations/opticalproperties.fcl
TPB time set to 0

OpDetReco/OpFlash/SBNDFlashFinder_module.cc
Removing if conditional which skips some OpChs when computing the Y, Z center of the OpFlash


ggamezdiego (approved):

ikatza:
Hey @fjnicolas since you're already cleansing code from various places, can you comment/remove these:
https://github.com/SBNSoftware/sbndcode/blob/753c1cb78403e10f0c305e67633a4e8036a4d551/sbndcode/OpDetReco/OpFlash/SBNDMCFlash_module.cc#L33

https://github.com/SBNSoftware/sbndcode/blob/753c1cb78403e10f0c305e67633a4e8036a4d551/sbndcode/OpDetReco/OpFlash/SBNDMCFlash_module.cc#L72

The pdsMap is not used in that module, so it's an unnecessary initialisation.

sbndcode/OpDetSim/DigiPMTSBNDAlg.cc
```diff
@@ -102,37 +126,70 @@ namespace opdet {
...
+    std::vector<double>& wave,
+    std::unordered_map<int, sim::SimPhotons>& DirectPhotonsMap,
+    std::unordered_map<int, sim::SimPhotons>& ReflectedPhotonsMap)
+  {
+
+    double ttsTime = 0;
+    double tphoton;
+    size_t timeBin;
+    double ttpb=0;
+    sim::SimPhotons auxphotons;
```
ikatza, line 163: It's better practice to initialise and define variables closer to where you need them. My suggestion is to define and initialise these inside each of the ifs below, the one for direct and reflected light.

sbndcode/OpDetSim/DigiPMTSBNDAlg.cc
```diff
@@ -102,37 +126,70 @@ namespace opdet {
...
+
+    //direct light
+    if(auto it{ DirectPhotonsMap.find(ch) }; it != std::end(DirectPhotonsMap) )
+    {auxphotons = it->second;}
+    for(size_t j = 0; j < auxphotons.size(); j++) { //auxphotons is direct light
+      if(CLHEP::RandFlat::shoot(fEngine, 1.0) < fQEDirect) {
+        if(fParams.TTS > 0.0) ttsTime = Transittimespread(fParams.TTS); //implementing transit time spread
```
ikatza, line 170:
```suggestion
        if(fParams.TTS > 0.0) double ttsTime = Transittimespread(fParams.TTS); // transit time spread
```

sbndcode/OpDetSim/DigiPMTSBNDAlg.cc
```diff
@@ -144,57 +201,95 @@ namespace opdet {
     double t_min,
     std::vector<double>& wave,
     int ch,
-    std::string pdtype,
-    std::unordered_map<int, sim::SimPhotonsLite>& directPhotonsOnPMTS)
+    std::string pdtype)
   {
     double mean_photons;
```
ikatza, line 206: Same here. Remove these declarations and declare and initialise below.

sbndcode/OpDetSim/DigiPMTSBNDAlg.cc
```diff
@@ -144,57 +201,95 @@ namespace opdet {
...
     }
 
+    //Adding noise and saturation
     if(fParams.PMTBaselineRMS > 0.0) AddLineNoise(wave);
     if(fParams.PMTDarkNoiseRate > 0.0) AddDarkNoise(wave);
     CreateSaturation(wave);
   }
 
 
+
```
ikatza, line 291: Remove extra space

sbndcode/OpDetSim/opDetDigitizerWorker.cc
```diff
@@ -223,19 +248,29 @@ void opdet::opDetDigitizerWorker::MakeWaveforms(opdet::DigiPMTSBNDAlg *pmtDigiti
...
                                           waveform,
                                           pdtype,
-                                          directPhotonsOnPMTS,
                                           startTime,
                                           fConfig.Nsamples);
           // including pre trigger window and transit time
           fWaveforms->at(ch) = raw::OpDetWaveform(fConfig.EnableWindow[0],
                                                   (unsigned int)ch,
                                                   waveform);
+          //std::cout<<"  Making Waveforms UncoatedPMTs  ChID="<<ch<<"  PDtype="<<pdtype<<"  Reflected?="<<Reflected<<std::endl;
```
ikatza, line 273: Either remove these couts or make them debug messages.

ikatza (commented):
Overall good. Only some minor observations about formatting that shouldn't affect the outcome.

sbndcode/OpDetReco/OpFlash/SBNDFlashFinder_module.cc
```diff
@@ -149,10 +149,6 @@ namespace opdet{
     double totalPE = 0.;
     double sumy = 0., sumz = 0., sumy2 = 0., sumz2 = 0.;
     for (unsigned int opch = 0; opch < pePerOpChannel.size(); opch++) {
-      if (opch > 31 && opch < 100){  //TO ADAPT FOR SBND USING THE MAP
```
marcodeltutto, line 152: Thank you! :)

sbndcode/OpDetSim/opDetDigitizerWorker.cc
```diff
@@ -202,15 +213,29 @@ void opdet::opDetDigitizerWorker::MakeWaveforms(opdet::DigiPMTSBNDAlg *pmtDigiti
...
                                                   (unsigned int)ch,
                                                   waveform);
-        }
+      	  //std::cout<<"  Making Waveforms ARAPUCA  ChID="<<ch<<"  PDtype="<<pdtype<<"  Reflected?="<<Reflected<<std::endl;
+      	}
       }
     }  //end loop on simphoton lite collections
+
+    //Constructing Waveforms for hybrid OpChannels (coated pmts)
+    for(auto ch : coatedpmts_todigitize){
```
marcodeltutto, line 222: This if fine for now, but in the future we may want to restructure this part a little bit so we don't single out coated pmts as a special case at the end. For example, we may want to have the PMT digitizer class to have methods like "AddPhotonsToWaveform" (instead of "ConstructWaveformLite") which takes in an existing waveforms and adds more photons to it.
But this is for the future, just writing it here so we can remember about it!

marcodeltutto (approved):
Very nice!

ascarff:
Is this PR ready to merge? I see @wesketchum has marked it as "done", but it still has outstanding comments from @ikatza and a missing review from @lpaulucc

wesketchum:
Yeah, sorry, I think my marking as "Done" was a mistake, and should now be "In progress" on the SBN 2020A project page. Am I right that this _must_ be resolved before we can start production? If so, we need to urgently get this complete.

ggamezdiego:
Hi, the bug has been solved and successfully tested. 
My understanding is that Iker's comments are basically "cosmetics", Is it correct Iker? And given the urgency of the situation, and as we expect to be moving soon to the new larg4 where the digitization module might be restructured, we decided to add those comments in further iterations. 
A review was asked from Laura as she is one of the original main authors of the module, in case she had a strong concern about the approach adopted to solve the bug. 
But as I said, the fixed module is doing now what is supposed to do.

wesketchum:
OK, thanks. If @lpaulucc can take a look that would be great, but it sounds to me that this is OK to be merged in ~tomorrow even without the extra set of eyes? Feel free to disagree.

sbndcode/LArSoftConfigurations/opticalproperties_sbnd.fcl
```diff
@@ -30,7 +30,10 @@ sbnd_opticalproperties: {
 
     LoadExtraMatProperties: true
 
-    TpbTimeConstant: 2.5
+    # TPB time set to zero in Geant4 beacuse it is unable to simulate the measured-TPB 
+    # time spectrum (Phys. Rev. C, 91,035503). This delay (i.e. the measured 4-exponential)
+    # is currently accounted at digitization level.
```
lpaulucc, line 35: Please check if the tpb time constant was actually included at digitization level in a correct manner for ALL kinds of detectors that see visible light. In order to have better timing characterization from the PDS, it would be better to have it at the propagation level but if this is the solution, this needs to be taken into account for the visible arapucas as well.

lpaulucc (changes requested):
Hi all,
Sorry for not replying before, I am trying to understand how gitHub works. I think the solution looks good. If it has been tested and it works, please go ahead with the merge as long as the correction for TPB time emission is also included for arapucas that detect visible light.
Laura

fjnicolas:
Hi, thanks for your comments @lpaulucc! I'll work on including the TPB time for the visible (x)arapucas today. It should be ready by tomorrow morning (Europe time)

fjnicolas:
Hi all, with the last commit the correction for the TPB time emission is also included for the visible (x)arapucas and the PR should be ready to merge.
WARNING!: the root file 'OpDetSim/digi_arapuca_sbnd.root' in sbnd_data must be updated. The updated root file can be found in ‘/sbnd/app/users/fnicolas/Data/digi_arapuca_sbnd.root’ (same file than before but including a std::vector with the TPB times)

lpaulucc (approved):
Now TPB time has been added to visArapucas in a different commit.

ascarff:
Ok, there is a new version of sbnd_data, v01_04_00, which has the updated version of `digi_arapuca_sbnd.root`. The next sbndcode release (v09_07_00) will point to this version.

Issue #47, opened 2020-10-16
Update SBND fcl files for 2020A production


wesketchum:
I have this from a message/quick look from Joseph. Don't know if it's accurate.

> 1. standard_g4_sbnd.fcl --> Electronlifetime: 3000
> 2. standard_detsim_sbnd.fcl --> Electronlifetime: 3000
> 3. standard_reco2_sbnd.fcl -->  Electronlifetime: 3000
> 4. Need to add SCE effects to SBND in-time generation
> 5. standard_reco2_sbnd.fcl --> "CalAreaConstants:  [ 0.02354, 0.02130, 0.02354 ]" Doesn't match DocDB 18991

I'd assign this to Mike Mooney, but don't see him as a member of SBNSoftware. So, sorry @jzennamo.

absolution1:
Those numbers do indeed describe the parameter state for sbndcode.  
The calo constants were last changed 4 years ago.

The lifetime should only need tweaking here: https://github.com/SBNSoftware/sbndcode/blob/develop/sbndcode/Utilities/detectorproperties_sbnd.fcl#L8

and the calo constants should only need tweaking here:
https://github.com/SBNSoftware/sbndcode/blob/develop/sbndcode/LArSoftConfigurations/calorimetry_sbnd.fcl#L9

jzennamo:
OK, I changed these to the values we agreed to. Question, how would we like these tested? 

I can check the fcl dumps for all the production fcls to check. Anything else?

wesketchum:
Checking fcl dumps sounds like a good first thing.

Calorimetry is maybe the end result to check then? Not sure what the best way to test this is otherwise.

@jzennamo, do you have a branch/pull request nearing ready that we can link? Just to keep track.

jzennamo:
I am writing up a quick script to dump all our production fcls so I can quickly check. 

I'll also have to work with @gputnam to address #4 above

jzennamo:
According to @gputnam the SBND space-charge configurations are incorrect, we should check that before closing this issue.

gputnam:
I think the space charge config is correct -- sorry -- it was a different piece of the config that I thought was wrong.

I think the space charge piece of the config looks ine.

jzennamo:
OK! Good to know, I added this feature branch that should gobble that up then:
https://github.com/SBNSoftware/sbndcode/compare/feature/zennamo_intime_with_sce

[...]

PR #51, opened 2020-10-19
[MCP2020A] Fix Intime Cosmic MC Generation

gputnam:
At some point in the past, the configurations I developed to generate intime-cosmic MC in SBND were changed (and then moved into the new JobConfigurations directory). Some of these updates changed the way the generation is implemented to something inconsistent with the original implementation.

This update, to the best of my ability, reverts those changes. I have dug around the ICARUS implementation (which was left intact) to find the configuration to use. I cannot completely vouch for this configs correctness relative to the original implementation, but it appears to run, etc.

### DIFF
--- a/sbndcode/JobConfigurations/standard/gen/corsika/prodcorsika_proton_intime_filter.fcl
+++ b/sbndcode/JobConfigurations/standard/gen/corsika/prodcorsika_proton_intime_filter.fcl
@@ -16,7 +16,9 @@ physics.producers.larg4intime: @local::sbnd_largeant
 physics.filters.GenInTimeSorter: @local::sbnd_filtergenintime
 physics.filters.timefilter: @local::sbnd_timefilterssimphotonlitetime 
 
-physics.simulate: [ generator, GenInTimeSorter, larg4intime, timefilter, rns ]
+physics.producers.corsika: @local::physics.producers.generator
+
+physics.simulate: [ corsika, GenInTimeSorter, larg4intime, timefilter, rns ]
 outputs.out1.SelectEvents: [ "simulate" ]
 
 physics.producers.larg4intime.KeepParticlesInVolumes: ["volCryostat", "volTaggerTopHigh", "volTaggerTopLow", "volTaggerSideLeft", "volTaggerSideRight", "volTaggerFaceFront", "volTaggerFaceBack", "volTaggerBot"] 
--- a/sbndcode/JobConfigurations/standard/gen/corsika/prodcorsika_proton_intime_filter.fcl
+++ b/sbndcode/JobConfigurations/standard/gen/corsika/prodcorsika_proton_intime_filter.fcl
@@ -15,7 +15,6 @@ services: {
 physics.producers.larg4intime: @local::sbnd_largeant
 physics.filters.GenInTimeSorter: @local::sbnd_filtergenintime
 physics.filters.timefilter: @local::sbnd_timefilterssimphotonlitetime 
-physics.filters.timefilter.Debug: true
 
 physics.simulate: [ generator, GenInTimeSorter, larg4intime, timefilter, rns ]
 outputs.out1.SelectEvents: [ "simulate" ]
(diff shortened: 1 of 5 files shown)


jzennamo:
Aw, in this should we include "enable_spacecharge_services_sbnd.fcl" in the in-time generation from https://github.com/SBNSoftware/sbndcode/pull/52 to address this problem #4 in https://github.com/SBNSoftware/sbndcode/issues/47 ?

gputnam:
@jzennamo I think(?) "enable_spacecharge_services_sbnd.fcl" is already included in the gen config: https://github.com/SBNSoftware/sbndcode/pull/51/files#diff-698c458592c0d659e0e6cf712f6601006f39c6019a8d2fc684a0c87e60fa2529.

sbndcode/JobConfigurations/standard/gen/corsika/prodcorsika_proton_intime_filter.fcl
```diff
@@ -1,11 +1,30 @@
...
+}
+
+physics.producers.larg4intime: @local::sbnd_largeant
+physics.filters.GenInTimeSorter: @local::sbnd_filtergenintime
+physics.filters.timefilter: @local::sbnd_timefilterssimphotonlitetime 
+
+physics.producers.corsika: @local::physics.producers.generator
+
+physics.simulate: [ corsika, GenInTimeSorter, larg4intime, timefilter, rns ]
 outputs.out1.SelectEvents: [ "simulate" ]
```
absolution1, line 22: Is there a particular need for gen/g4 to be run in the same fcl job?

sbndcode/JobConfigurations/standard/gen/corsika/prodcorsika_proton_intime_filter.fcl
```diff
@@ -1,11 +1,30 @@
...
+
+physics.producers.larg4intime.KeepParticlesInVolumes: ["volCryostat", "volTaggerTopHigh", "volTaggerTopLow", "volTaggerSideLeft", "volTaggerSideRight", "volTaggerFaceFront", "volTaggerFaceBack", "volTaggerBot"] 
+physics.producers.larg4intime.InputLabels: [ "GenInTimeSorter:intime" ]
+
+process_name: CosmicsCorsikaCMCGenAndG4InTime
+
+#include "3drift_services_sbnd.fcl"
+#include "enable_spacecharge_services_sbnd.fcl"
```
absolution1, line 30: if 3drift and space charge are needed then the fcl file name should probably reflect that.

absolution1 (commented):
Looks nice @gputnam 
I had a couple of minor comments/questions.

sbndcode/JobConfigurations/standard/gen/corsika/prodcorsika_proton_intime_filter.fcl
gputnam: Good point!

sbndcode/JobConfigurations/standard/gen/corsika/prodcorsika_proton_intime_filter.fcl
gputnam: We can discuss what's the best strategy -- my reasoning is:

The second stage of the filtering requires a check on the number of photons hitting the optical detectors. So we need to run G4 before that stage. However, it would be very slow to run G4 on 3-drift-windows of cosmics just to get filtered out. So the strategy is, just run G4 on the cosmics that might have in-time light. This makes the filtering much faster.

So we run intime G4 -> photon filter -> out-time G4. I put the intime G4 / photon filter in the gen stage for two reasons:
-I wanted all the filtering to all be done in one stage to ease production somewhat
-I'm not sure about running two instances of G4 in the same job (since running the Geant module relies on a lot of state that is in a single place per-process)

Let me know what you think!

absolution1: This all sounds annoyingly fiddly haha
OK! I totally understand the rationale and it sounds like you've been careful with this decision.

It's obviously always good to try and keep the same structure throughout the various production workflows to ease the cognitive load when reading the FCLs.  However, when there's a good reason to diverge, you should diverge and it sounds like this is a good reason!

I guess the other alternative would be including an extra stage in the production workflow --or-- running two `lar` commands (one which runs `intime` and one which runs `outoftime` during the G4 stage for this particular campaign.

All approaches (including what you've done) are totally valid and I'd leave the decision to you and the other organisers.

(I also wouldn't trust legacy LArG4 to handle the seed state in this case!)

sbndcode/JobConfigurations/standard/gen/corsika/prodcorsika_proton_intime_filter.fcl
gputnam: Fixed.

PR #52, opened 2020-10-19
[MCP2020A] Space Charge + 3 Drift Window Configs

gputnam:
For all of the event selection samples we need space charge and 3 drift windows both enabled.

I have added in a base config to enable the space charge services. I have also added in every config for g4/detsim/reco needed with both settings on.


wesketchum:
Is any of this duplicated with #48 ?

gputnam:
uhhh yeah looks like one of the files is -- reco2_3drift_windows_sce.fcl

sbndcode/JobConfigurations/standard/detsim/detsim_3drift_windows_sce.fcl
```diff
@@ -0,0 +1,4 @@
+#include "standard_detsim_sbnd.fcl"
+
+#include "3drift_services_sbnd.fcl"
+#include "enable_spacecharge_services_sbnd.fcl"
```
absolution1, line 4: Probably a dumb Q but does anything in the detsim chain need access to the SCE service?

sbndcode/JobConfigurations/standard/reco/reco1_3drift_windows_sce.fcl
```diff
@@ -0,0 +1,4 @@
+#include "standard_reco1_sbnd.fcl"
+
+#include "3drift_services_sbnd.fcl"
+#include "enable_spacecharge_services_sbnd.fcl"
```
absolution1, line 4: Is the SCE service needed for reco1?

absolution1 (commented):
Thanks @gputnam looks nice!
Had a couple of minor questions.

absolution1:
> uhhh yeah looks like one of the files is -- reco2_3drift_windows_sce.fcl

Hey, I did warn you that I was committing the reco2 FCLS.  I figured the PR had already been merged.
I guess we have to figure out the ordering.
I guess that @ascarff could merge #48 and then this PR is updated to keep the history in order.

sbndcode/JobConfigurations/standard/detsim/detsim_3drift_windows_sce.fcl
gputnam: I don't know, but I added it in to be on the safe side. I figured there was no downside in being consistent.

sbndcode/JobConfigurations/standard/reco/reco1_3drift_windows_sce.fcl
gputnam: Same as for detsim.

gputnam:
@absolution1 yeah that sounds good to me!

sbndcode/JobConfigurations/standard/detsim/detsim_3drift_windows_sce.fcl
absolution1: Ah ok, sure. 
There is a bit of a downside here, in that there's extra points of maintenance.
Personally, I would say that, if `gen -> g4_sce -> detsim -> reco1 -> reco2_sce` runs/works then that would probably be better.  But, I'd obvs defer the decision to you / @jzennamo / @wesketchum

PR #53, opened 2020-10-19
[MCP2020A] Enable Flash Matching in Reconstruction

gputnam:
This pull request enables the Flash-Matching module to be run in SBND reconstruction.

It also updates CMake to properly build the sbndPDMapAlg as a tool (which appears to be intended, but is not the way it works currently).


wesketchum:
Is this pending the changes in LArSoft/larsim#44 as well?

gputnam:
No -- this turns on Iker's flash matching.

My understanding was that Marco's flash matching is not ready and is not going to be in the MC production.

marcodeltutto:
Hi Gray, it is ready and there are two PRs (one in sbncode and one in sbndcode) which add the code and and enable it in the reco fcl file. This will have to go in the MC production as well.

gputnam:
Ah, ok. In any case, this pull request is independent from that stuff.

Issue #63, opened 2020-11-02
[MCP2020A] reco2_sce fcls do not run


etyley:
`reco2_sc2.fcl` and `reco2_3drift_windows_sce.fcl` do not run in version v09_09_00 and are needed for production. When trying to  run I get the following error:
```%MSG-s ArtException: 
cet::exception caught in art
---- Configuration BEGIN
  The following error was encountered while processing a path configuration:
  Entry flashmatch in path reco2_sce does not have a module configuration.
---- Configuration END
%MSG```

PR #64, opened 2020-11-02
[MCP2020A] Fix reco2_sce fcls 

etyley:
Fixes issue #63 by updating the naming of `flashmatch` to `opt0finder`. I think the name of the module was updated previously but the change was not propagated to the `sce` fcls.

I also took the opportunity to add the flash matching to SCE corrected Pandora, in line with what is done for `fmatch` and `fmatchSCE`

@marcodeltutto should check this is correct before it is merged.


marcodeltutto (approved):
Looks perfect, thank you!

PR #70, opened 2020-11-17
Simple Flash Matching migration and update

ikatza:
Removing the code that was hosted here for the Simple Flash Matching as it's going to `sbncode` now.

I've decided to keep the fcl files that control SBND in here.

This should be merged along with
https://github.com/SBNSoftware/sbncode/pull/59


ikatza:
Besides the PR on `sbncode` that I link above, this PR also needs an update of the metrics file that's on `sbnd_data`. 

I have asked @ascarff for help to update that product too.

ascarff:
sbnd_data v01_05_00 has the new files, I'll include that in the next release with this PR. (This comment is mainly to help me remember...)

PR #73, opened 2020-12-11
Update Pandora config xml to use recursive mop up

etyley:
Enable recursive shower growing in Pandora, see DocDB 19931 for performance details.


absolution1 (approved):
Cracking stuff.

PR #90, opened 2021-03-22
Remove TFileMetadataSBND and FileCatalogMetadataSBND

cjbackhouse:
They are moving to "SBN" versions in sbncode in the branch of the same name there.


etyley:
I pushed a change to the default fcl configuration used for configuring production jobs. I am running some tests now to check that everything works as expected still.

etyley (approved):
I have tested this now and it all runs as expected

PR #93, opened 2021-03-24
Fixed optrigger alg to produce uniformly-sized waveforms, cleaned up a few snippets, and set readout length to 12us

wforeman:
While testing changes to the TPC readout I noticed the optical trigger alg was creating waveforms of various lengths instead of a uniform size of 4us. I restructured this part of the code so that triggered waveforms are all of a single size. Other than that, the functionality should be the same. My understanding is eventually we'll want to change this alg so that a threshold met in one detector triggers *all* of them to read out at the same time, but that goes beyond the scope here.

The trigger enable window is now set to start accepting triggers a full drift period (~1.3ms) before the start of the TPC readout, which is required in order to be able to match to every track in the TPC.

Also made minor changes to a few other places in the code and added "xarapuca_vuv" and "xarapuca_vis" to the channel mask (since I believe these are now the names of those guys?).

Finally, the default readout length was increased from 4us to 12us, since the 4us number is outdated.

### DIFF
--- a/sbndcode/OpDetSim/opDetSBNDTriggerAlg.cc
+++ b/sbndcode/OpDetSim/opDetSBNDTriggerAlg.cc
@@ -21,7 +21,6 @@ struct TriggerPrimitive {
   raw::TimeStamp_t start;
   raw::TimeStamp_t finish;
   raw::Channel_t channel;
-
 };
 
 // Local static functions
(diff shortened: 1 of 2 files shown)


pgreen135 (approved):

ikatza:
I was telling Will that I didn't had time this week to review this. Guess won't be necessary then.

pgreen135:
Hi @ikatza, I was not aware you were intending to review this. If you would like I can revert the merge

ikatza:
It's fine. I trust that you didn't find anything wrong and you deemed it good to be merged.

I was just commenting Will that from the changes he's making, it's clear we haven't updated this in quite a while. So I wanted to have more time to test things and possibly doing some more updating.

PR #95, opened 2021-03-31
New Flux Files Config H

### LINKED ISSUE #79: Beam spot is in the wrong place
Beam axis should be at x=-74 cm in the larsoft SBND detector coordinate system based on the latest survey measurements.
Currently it is at x=+50cm

Cross check of this number is ongoing with the data taken with the CRT panels in the pit.

marcodeltutto:
This PR add the necessary fhicl configurations to use the newly produced flux files. It fixes #79. The current flux files (config F) have the beam spot at a 45.7 cm offset in detector coordinates. According to the current design this offset should instead be -73.78 cm, as mentioned in #79 and described in docdb 20891. The new files (called config H) were produced by Zarko and have the corrected -73.78 cm offset.

These new flux files config H are in `/cvmfs/sbnd.osgstorage.org/pnfs/fnal.gov/usr/sbnd/persistent/stash/fluxFiles/bnb/BooNEtoGSimple/configH-v1/march2021/neutrinoMode/`.

Validation plots are on [docdb 21351](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=21351).

The wiki page has been updated: https://sbnsoftware.github.io/sbndcode_wiki/The_SBND_flux_files.html.

Here are more details about this PR:
- Flux config H added to `LArSoftConfigurations/genie_sbnd.fcl`.
- All the fcls with the different flux configs have been moved to `genie/other_flux_config/flux_config*/` and `overlay/other_flux_config/flux_config*/`.
- The directories `genie/` and `overlay/` now have default fcls files to use for event generation (that don't have gsimple_config*_v1 in the name):

```
ls genie/
prodgenie_nu_spill_cryostat_sbnd.fcl
prodgenie_nu_spill_tpc_sbnd.fcl
prodgenie_nu_singleinteraction_cryostat_sbnd.fcl
prodgenie_nu_singleinteraction_tpc_sbnd.fcl
prodgenie_nu_singleinteraction_tpc_ccpi0filtered_sbnd.fcl 
prodgenie_intrnue_spill_tpc_sbnd.fcl
prodgenie_intrnue_singleinteraction_tpc_sbnd.fcl
prodgenie_nu_singleinteraction_tpc_rotatedbuckets_sbnd.fcl
prodgenie_nu_spill_world_sbnd.fcl
other_flux_configs/

ls overlay/
prodoverlay_corsika_cosmics_cmc_genie_nu_spill_cryostat_sbnd.fcl
prodoverlay_corsika_cosmics_cmc_genie_nu_spill_tpc_sbnd.fcl
prodoverlay_corsika_cosmics_proton_genie_intrnue_spill_tpc_sbnd.fcl
prodoverlay_corsika_cosmics_proton_genie_nu_spill_cryostat_sbnd.fcl
prodoverlay_corsika_cosmics_proton_genie_nu_spill_tpc_sbnd.fcl
other_flux_configs/
```

@etyley Fhicl files that we run in production for GENIE only and GENIE+Corsika should be updated with the following recipe:
prodgenie_nu_spill_tpc_**gsimple-configf-v1**.fcl -> prodgenie_nu_spill_tpc_**sbnd**.fcl


sbndcode/JobConfigurations/standard/gen/genie/default/prodgenie_intrnue_singleinteraction_tpc_sbnd.fcl
```diff
@@ -0,0 +1,5 @@
+# Simulates GENIE nue and anue neutrino interactions from the BNB beam 
+# forcing one interaction per event, inside the TPC volume
+# (with 10 cm padding on each side).
+
+#include "prodgenie_intrnue_singleinteraction_tpc_gsimple-configh-v1.fcl"
```
etyley, line 5: I would have expected this dependency to be the other way around: that the specific flux versions of the fcls would simply inherit from the default versions and only override the flux config. The default flux configuration could then be set in a single upstream fcl (presumably `genie_sbnd.fcl`). This current implemntation requires all of these fcls to be changed whenever the flux is updated and we have run into issues with this sort of thing in the past many times which is why I would advocate for a single point of maintenance here.

marcodeltutto: I agree with you, I assumed there was a reason to have all these fcl files for all the different flux configurations, and so my approach was to keep all these files, add the ones for config H, and then I added the default ones for simplicity. But if you confirm that there is no need to have all the fcls for all the configurations, I'll be happy to change the way you propose, as much simpler to maintain.

etyley: I think having all of the files for posterity and reproducibility is not a bad thing. The approach I would have taken is to create some fcls that just override the flux config (e.g. https://github.com/SBNSoftware/sbndcode/blob/develop/sbndcode/LArSoftConfigurations/prodgenie_nue.fcl). Then your flux config files could just be as simple as:
```
#include <some default fcl>

#include <flux config fcl>
```
Although maybe this is a larger discussion about organisational changes than this pull request

marcodeltutto: Ok, I like the fcls that override. I'll take a look at this later today update this PR. Thanks for the feedback!

marcodeltutto: I added more fcl-override files for genie, and called all of them `set_genie_*.fcl` (ex `set_genie_nue.fcl`) and used them to set the flux configuration.

The main fcl files that should be used as have the most recent flux files are in `JobConfigurations/standard/gen/genie`, while I put all the other flux configurations in a sub-directory called `other_flux_config`. Same is true for `JobConfigurations/standard/gen/overlay`.

pgreen135:
@marcodeltutto is this ready to be merged?  @absolution1 did you want to review it first? I will be doing a new release tomorrow, and can merge this in if it is ready

pgreen135:
trigger build

marcodeltutto:
I think since @etyley looked at this it is fine to merge. Happy to amend things later if @absolution1 has further comments!

pgreen135:
@vitodb the FNALbuild bot is still spamming this, not sure what is going on

PR #97, opened 2021-04-15
TPC readout window reduction

wforeman:
This PR includes all the changes related to setting the new default TPC readout time to 1.3 ms with a 0.2 ms front/back porch (see [sbn-doc-21573](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=21573)). Specific changes outlined below. The feature branch name is misleading, this is actually a ~15% porch.

/Utilities:
- Set TPC readout window length, trigger offset, default beam/trigger time, and G4 reference times in detectorclocks_sbnd.fcl and detectorproperties.fcl

/LArSoftConfigurations:
- Corsika/Cry generation time parameters updated, each sharing a common fcl parameterset contained in cosmic_common_sbnd.fcl
- Added "1drift_services_sbnd.fcl", analogous to "3drift_services_sbnd.fcl" but for a single drift window with no front or back porch (how prodsingle used to work)
- Updated T0 in singles_sbnd.fcl to match new default G4 reference time
- Updated event display time samples to new readout window length
- Added "drop_raw_products.fcl" which drops heavy raw products. I've found it useful for drastically reducing disk space usage on PNFS during the readout studies, but we can drop this from the PR if others think it shouldn't be here.

/JobConfigurations:
- Added "3drift" versions of the prodoverlay files (both BNB nu flux and intrnue). The new default will be the 15% porch, but this file is here in case anybody wants to run in the old 3-drift-window configuration for studies.
- Updated T0 in all the prodsingle scripts that include it (for scripts only simulating 1 particle, this value will be inherited from singles_sbnd.fcl)
- Added SCE versions of g4, detsim, reco1

/AnalysisTree:
- Added PFP and slice info to tree
- Added more hit variables to tree: width, mcid, frac (fraction of hit energy from 'mcid' particle), true energy, true number electrons reaching wires, and "reco" number of electrons (scaling the hit area with the calibration constants via CaloAlg)
- Importing the calorimetry alg in order to do the ADC-to-electron conversion on hits
- Helper functions to consolidate unsightly truth-matching procedures 
- Included electron lifetime into tree
- Added extra Geant variables to tree: particle's deposited energy (DepEnergy) and particle's number of electrons arriving at wires (NumElectrons)
- Added total deposited energy and total number of drifted electrons in event to tree
- Promoted ParticleInventoryService and BackTrackerService instances to be class-wide objects, so we don't have to keep re-declaring them all over the place
- Tidied up some code blocks and fixed indentation for clarity
- Calculations of slice purity and completeness for neutrino-tagged slices


jzennamo:
@wforeman, Could you highlight the fcl chain that should be run to get the new "standard" 0.15-1-0.15 setup for the BNB+cosmic and in-time cosmic generations? [I am still reviewing but I'll pull this down, do a build test, and do a fhicl-dump review of these workflows]

sbndcode/JobConfigurations/standard/gen/single/prodsingle_electron_pi+_bnblike_forward.fcl
```diff
@@ -17,7 +17,7 @@ physics.producers.generator.Y0: [0, 0]
 physics.producers.generator.SigmaY: [ 0 ]
 physics.producers.generator.Z0: [250, 250]
 physics.producers.generator.SigmaZ: [ 0 ]
-physics.producers.generator.T0: [0, 0]
```
jzennamo, line 20: This file looks like it creates two particles should this line be `[1700, 1700]`

sbndcode/LArSoftConfigurations/cosmic_common_sbnd.fcl
```diff
@@ -0,0 +1,12 @@
+#
+# Cosmic generation configurations shared between Corsika and Cry
+BEGIN_PROLOG
+
+sbnd_cosmic_common: {
+  SampleTime:       3.2e-3  #  0.2 ms (G4 rise time) + 1.3 ms (1 full drift window) + 1.7 ms (readout) @ 500 V/cm, 154.4 cm/ms
+  TimeOffset:      -1.7e-3  # -0.2 ms (G4 rise time) - 1.3 ms (1 full drift window) - 0.2 ms (front porch)
```
jzennamo, line 7: Just to make sure I understand, we need the comics to start `0.2 ms (rise time) + 1.3 ms (1 full drift window)` before the trigger, is that what is in here?

jzennamo (commented):
I had these two comments let me know what you think

sbndcode/JobConfigurations/standard/gen/single/prodsingle_electron_pi+_bnblike_forward.fcl
wforeman: Oh, good catch, thanks. Maybe I had assumed the "pad vectors" option would work, but always better to be explicit.

sbndcode/LArSoftConfigurations/cosmic_common_sbnd.fcl
wforeman: Not quite; we need cosmics to start 1 full drift window before the _readout_, not the trigger. Readout starts at -0.2 ms relative to the trigger. So that's `0.2 ms (rise time) + 1.3 ms (full drift) + 0.2 ms (front porch) = 1.7 ms prior to trigger`

jzennamo: Perfect! Thank you that makes sense.

wforeman:
RE Joseph's comment: since these settings are now default, the standard fcl workflow will use the new 15%-100%-15% configuration. The XML I used to do the final test is `/sbnd/app/users/wforeman/sbndcode_readout/xml/finaltest.xml`, which uses the chain below:

`prodoverlay_corsika_cosmics_proton_genie_nu_spill_tpc_sbnd.fcl`
`standard_g4_sbnd.fcl`
`standard_detsim_sbnd.fcl`
`standard_reco1_sbnd.fcl`
`standard_reco2_sbnd.fcl`
`(drop_raw_products.fcl)`

The last fcl is just there to conserve disk space and speed up file transfers from the grid nodes during the test.

sbndcode/JobConfigurations/standard/gen/single/prodsingle_electron_pi+_bnblike_forward.fcl
jzennamo: OK, let me know when you've resolved this and I'll close this issue, thanks!

absolution1: Hi there @wforeman 
Is it possible to slot in a fcl parameter rather than a hardcoded param here?  `@local::something_something` or `@sequence::something_something` or `@table::something_something` --might-- work here

wforeman:
I didn't test this using in-time cosmics, but the fcl chain would be the same except for the gen stage:
`prodcorsika_proton_intime_filter.fcl`

For in-time w/SCE, the chain would be:
`prodcorsika_proton_intime_filter.fcl`
`g4_sce_simphotontime_filter.fcl` (just added)
`detsim_sce.fcl`
`reco1_sce.fcl`
`reco2_sce.fcl`

For standard BNB w/SCE, the chain is:
`prodoverlay_corsika_cosmics_proton_genie_nu_spill_tpc_sbnd.fcl`
`g4_sce.fcl`
`detsim_sce.fcl`
`reco1_sce.fcl`
`reco2_sce.fcl`

sbndcode/JobConfigurations/standard/gen/single/prodsingle_electron_pi+_bnblike_forward.fcl
wforeman: I've been following the convention in all the existing prodsingle scripts, which all explicitly list out the T0 array for multi-particle guns. It's a great suggestion to clean those up and consolidate parameters like you suggest, but maybe that's a matter for a separate pull request?

marcodeltutto:
Thanks Will! I'll take a look this afternoon, I am getting your branch built to check a few things and I'll also try to visualize some events with TITUS to double check and to make sure there are no other changes needed in TITUS itself.

sbndcode/JobConfigurations/standard/gen/single/prodsingle_electron_pi+_bnblike_forward.fcl
absolution1: Are there many individual FCLs that explicitly set the T0? ( I'm hoping that there's a whole load of inheritance instead)
If there's only one or two then I'd say make the change now.  If there's a lot then, yeah, something for later.  If it's the latter, would you mind filing a GitHub issue so we don't forget?

wforeman: There are 9 files that generate doubles (T0: [1700, 1700]) and 2 that generate quadruples (T0: [1700, 1700, 1700, 1700]). All the rest, which just generate 1 particle, now inherit their T0 from singles_sbnd.fcl as of this PR.

absolution1: Thanks for checking.  A GitHub issue and a separate PR is the way forward.  In your copious spare time can you flag the GitHub issue?

jzennamo:
OK, I looked through the fhicl-dumps for the new "standard" fcls and things look as I expected, once Marco signs off I am happy. Thanks!

marcodeltutto (approved):
Looks good to me! I have also dumped the fcl configs and checked they have the right settings. I run some test jobs, all successful, and the events look as expected. One example of such events visualized with TITUS is attached here.
Screen Shot 2021-04-16 at 3 12 21 PM

pgreen135:
trigger build

pgreen135 (approved):

pgreen135:
looks like we are still having some issues with the automated CI triggers when PRs are merged; apologies for the spam, we will try to fix this asap @henrylay97 @vitodb

PR #100, opened 2021-04-22
Updates and fixes to optical trigger alg

wforeman:
This PR includes fixes to the optical triggering alg, based on conversations with Gray Putnam and Michelle Stancari.

Changes:
- Changed name of `TriggerEnableWindowIsTPCReadoutWindow` to `TriggerEnableWindowOneDriftBeforeTPCReadout` to better reflect the actual behavior of the alg. Still a mouthful, but hopefully less confusing.
- Added drift period as a fcl parameter
- Beam trigger is now implemented in the `FindTriggers` function along with the usual threshold-based triggers, instead of in `ApplyTriggers` (which _consumes_ the triggers found from `FindTriggers`)
- Added option to toggle on/off trigger overlap, partially undoing some misguided changes I made previously where all waveforms were made with uniform size. A uniform readout size is now achieved only when `AllowTriggerOverlap = false`.
- Restored distinction between beam vs non-beam readout window sizes
- Implemented the trigger holdoff functionality. This parameter existed before, but wasn't yet being used. Default behavior is unchanged (`TriggerHoldoff = 0`)
- `ReadoutWindowPreTriggerBeam` is now assigned the right value

I've tested this locally and confirmed it makes waveforms at the expected trigger locations, and that their size varies as expected based on the `AllowTriggerOverlap` setting. I also tested the trigger holdoff, which works nicely at preventing the occurrence of multiple triggers all happening within close proximity. It's set to 0 by default (should consult CAEN manual to set a more realistic value eventually).


sbndcode/OpDetSim/opDetSBNDTriggerAlg.cc
```diff
@@ -245,28 +262,13 @@ void opDetSBNDTriggerAlg::MergeTriggerLocations() {
...
-    double drift  = 1300; // 1.3 ms
-    double readout= detProp.ReadOutWindowSize() * clockData.TPCClock().TickPeriod();
-    start = clockData.TriggerOffsetTPC() - drift;
-    end = clockData.TriggerOffsetTPC() + readout;
+
+  // Enable triggers starting 1 drift period prior to the start of TPC readout,
+  // and continuing over the full range of the TPC readout window
+  if (fConfig.TriggerEnableWindowOneDriftBeforeTPCReadout()) {
+    double tpcReadoutWindow = detProp.ReadOutWindowSize() * clockData.TPCClock().TickPeriod();
+    start = clockData.TriggerOffsetTPC() - fConfig.DriftPeriod() - 1; /* Give 1us of wiggle room */
```
gputnam, line 270: It looks like you've added in a configuration to set the BeamTriggerTime. Shouldn't that be included here for instances where it is not equal to 0?

sbndcode/OpDetSim/opDetSBNDTriggerAlg.cc
```diff
@@ -324,65 +326,94 @@ double opDetSBNDTriggerAlg::ReadoutWindowPostTriggerBeam(raw::Channel_t channel)
...
+  unsigned  min_ro_samples = ro_samples;
 
   for(size_t i=0; i<adcs.size(); i++){
     double time = tick_to_timestamp(clockData, waveform.TimeStamp(), i);
 
     // if we're out of triggers or nearing the end of the waveform, break
     if( trigger_i >= trigger_times.size() ) break;
-    if( time >= end_time - ro_length ) break;
-    
+    if( i >= adcs.size()-1-min_ro_samples ) break;
```
gputnam, line 370: I can see the effort to include both the extend and include behaviors of the readout. It is a little hard for me to parse the logic, and the code is quite different now, so I will assume that both options have been tested.

One question (which is a bit of a nitpick) -- there is now this "Deadtime" configuration option. Can't you just implement the ignore behavior by setting the deadtime to the length of the readout? Then it avoids the need for special case logic here. But just a thought.

gputnam (commented):
See comments.

sbndcode/OpDetSim/opDetSBNDTriggerAlg.cc
wforeman: Good point! I'll add that.

I followed the previous convention here, which also ignored BeamTriggerTime when setting the enable window start/end. As far as I can tell, by the detsim stage, the timing of everything should be shifted such that beam = 0. But good to include it here for consistency, in case that ever isn't true for some reason...

sbndcode/OpDetSim/opDetSBNDTriggerAlg.cc
wforeman: I did test both options (AllowOverlap = true/false) over a number of events, and the behavior was as we expect, with readout sizes either expanding as needed to encompass overlapping triggers, or remaining fixed. Also tested several different pre/post-trigger readout lengths for both beam and off-beam.

The problem with setting the deadtime (TriggerHoldoff) to be the RO window length is that it would prevent overlapping triggers. Effectively, `TriggerHoldoff = RO length` has the same effect as `AllowTriggerOverlap = false`. So I think we need this special-case logic here, to prevent a readout from starting too close to the end of the available time-series.

gputnam:
Thanks Will! That addresses the questions I had.

pgreen135:
trigger build

pgreen135:
@mstancar did you want to review this before it's merged? if you are happy with it, I will aim to include it in this weeks release (Friday)

mstancar:
I am happy with the concept, don't have time to check the code for issues

________________________________

pgreen135:
@wforeman I see from the CIs above that there is a reduction in the number of raw::OpDetWaveform. This seems unsurprising given the changes, but could you confirm that this is expected?

395: 18c18
396: < DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 369
397: ---
398: > DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 244

wforeman:
@pgreen135, I think this makes sense. Now that readouts are allowed to vary in size once again, there will be fewer of them overall, since many long-ish readouts will replace previous instances where multiple fixed-size readouts occurred back to back.

PR #102, opened 2021-04-28
Add 1-drift services file for backwards compatibility

wforeman:
Just adding a small fhicl file to use in case anybody is processing older MC, made prior to the default drift window and readout length changes. We'll probably want to eventually remove this file down the line.


pgreen135:
trigger build

PR #107, opened 2021-05-12
Reduce verbosity of SBNDOpT0Finder

ikatza:
In my opinion is fairly normal to not have good flashes from one of the TPCs. Hence reducing the warning to an info.


marcodeltutto (approved):
Makes total sense, thank you!

PR #108, opened 2021-05-21
Filters for commissioning studies

marcodeltutto:
Integrating work done by @mstancar:
- Gen and LArG4 filters to select cosmics crossing certain CRT planes,
- adds fcls to filter front/back and east/west CRT muons
- adds a detsim fcl for commissioning studies
- cleans up the HitDumper module and add POT information into the output tree for neutrino studies


pgreen135:
trigger build

mstancar (approved):
Can you increase the MaxHits Variable in the hitdumper_module.cc ?
maybe change the CRT -> 5000, the PMTs and TPC to 100,000? Based on what Erin was seeing?

marcodeltutto:
@mstancar Done!
- maximum number of hits for wire and optical have been increased to 100k, and to 5k for the CRT.
- I took the occasion to change from arrays to std::vectors, so we can now change the maximum numbers via fcl, and it also uses less memory at run time (~7GB with arrays, ~2GB with vectors).
- Based on the discussion with Erin, I am now saving all the ophits (not only from coated pmts as it was before), so we can select them later

pgreen135:
trigger build

sbndcode/Commissioning/HitDumper_module.cc
```diff
@@ -377,23 +393,23 @@ void Hitdumper::analyze(const art::Event& evt)
...
-    _hit_cryostat[i]    = wireid.Cryostat;
-    _hit_tpc[i]         = wireid.TPC;
-    _hit_plane[i]       = wireid.Plane;
-    _hit_wire[i]        = wireid.Wire;
-    _hit_channel[i]     = hitlist[i]->Channel();
+    _hit_cryostat.push_back(wireid.Cryostat);
+    _hit_tpc.push_back(wireid.TPC);
+    _hit_plane.push_back(wireid.Plane);
+    _hit_wire.push_back(wireid.Wire);
+    _hit_channel.push_back(hitlist[i]->Channel());
```
pgreen135, line 407: @marcodeltutto with the change to using vectors it looks like you are now repeatedly resizing vectors when doing `push_back` here and everywhere else. Could you instead reserve the required size e.g. `_hit_cryostat.reserve(_nhits);` for each of these beforehand?

marcodeltutto: @pgreen135 Yes, good point, thanks! Done with commits c699437 and 0d49d7d.

PR #110, opened 2021-06-09
Hybrid model for Light simulation UGR

claudiaalvgar:
Pull Request introducing the configuration fcl and the g4 fcl file needed for using the new Hybrid model for light propagation for the new larg4 in LArSoft. The idea of this model is to use the Semi-Analytic model for propagating photons inside the active volume and a scaled-down Optical-Library for the LAr volume outside the TPC. This way, we are able to have a complete light simulation having into account the whole cryostat volume.
- For using this model we need to modify the PDFastSim_sbnd.fcl, including: the configuration of the Semianalytic module and the OpLibrary module, so that we are able to run both modules (the first one inside the active volume and the second one outside the active volume).
- We have also included the g4 fcl file for running the Hybrid Model which is called: OpHybrid_g4_refactored_sbnd.fcl. This fcl does the propagation of photons using the Semianalytic model if the scintillation point is generated inside the active volume (volTPCActive) and it runs the OpLibrary module if the scintillation point is outside the active volume.

### DIFF
--- a/sbndcode/JobConfigurations/standard/g4/OpHybrid_g4_refactored_sbnd.fcl
+++ b/sbndcode/JobConfigurations/standard/g4/OpHybrid_g4_refactored_sbnd.fcl
@@ -53,11 +53,12 @@ services:
        category      : "world"
        #List of volumes where Ionization/scintillation are simulated (i.e. material = "LAr")
        #Volumes used by the Hybrid Model
+       #Warning: make sure that names of LAr volumes are correct regarding the geometry used
        volumeNames   : ["volTPCActive", "volCryostat", "volTPCPlaneVert","volXArapuca","volPMT","volFieldCage","volPDSstructure"]
-       # Corresponding stepLimits in mm for the volumes in the volumeNames list
+       #Corresponding stepLimits in mm for the volumes in the volumeNames list
        stepLimits    : [0.3, 0.3, 0.3, 0.3, 0.3, 0.3, 0.3] 
-      #gdmlFileName_ : "sbnd_v02_00_nowires.gdml"
-       gdmlFileName_ : "sbnd_v02_00.gdml"
+       gdmlFileName_ : "sbnd_v02_00_nowires.gdml"
+      
       }
     
     MCTruthEventAction:  
@@ -177,12 +178,9 @@ outputs:
     }
 }
 
-services.LArPropertiesService.ScintPreScale: 1
 
 services.LArG4Parameters.UseLitePhotons: true   # true = produce SimPhotonsLite and OpDetBacktrackerRecord, false = produce SimPhotons
+services.LArG4Parameters.UseModLarqlRecomb: true
 
- services.LArG4Parameters.UseModLarqlRecomb: false
-
-services.Geometry.GDML: "sbnd_v02_00.gdml"
 services.AuxDetGeometry.GDML: "sbnd_v02_00.gdml"
 services.Geometry.DisableWiresInG4: false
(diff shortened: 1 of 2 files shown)


pgreen135:
trigger build LArSoft/larsim#68

sbndcode/JobConfigurations/standard/g4/OpHybrid_g4_refactored_sbnd.fcl
```diff
@@ -0,0 +1,188 @@
...
+    }
+}
+
+services.LArPropertiesService.ScintPreScale: 1
+
+services.LArG4Parameters.UseLitePhotons: true   # true = produce SimPhotonsLite and OpDetBacktrackerRecord, false = produce SimPhotons
+
+ services.LArG4Parameters.UseModLarqlRecomb: false
+
+services.Geometry.GDML: "sbnd_v02_00.gdml"
```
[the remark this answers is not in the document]
pgreen135: also this geometry is not actually available in sbndcode, so this isn't going to work @ggamezdiego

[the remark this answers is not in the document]
ggamezdiego: The override mentioned by Iker is actually not needed. And regarding the version of the geometry, we know it is not available, but we hope it will be very soon.  The reason for keeping that version there is because the new scaled-down Optical-Library has been generated using that version.

claudiaalvgar:
Thank you for the comments, I have uploaded an updated version of the g4 fcl file for the Hybrid Model

sbndcode/JobConfigurations/standard/g4/OpHybrid_g4_refactored_sbnd.fcl
```diff
@@ -0,0 +1,186 @@
...
+        outputCommands:   [ "keep *"]
+        compressionLevel: 1 #zlib argument (0-9) 
+    }
+}
+
+
+services.LArG4Parameters.UseLitePhotons: true   # true = produce SimPhotonsLite and OpDetBacktrackerRecord, false = produce SimPhotons
+services.LArG4Parameters.UseModLarqlRecomb: true
+
+services.AuxDetGeometry.GDML: "sbnd_v02_00.gdml"
```
ikatza, line 185: There remains this override.

sbndcode/JobConfigurations/standard/g4/OpHybrid_g4_refactored_sbnd.fcl
```diff
@@ -0,0 +1,186 @@
...
+    LArG4Detector : 
+    {
+       category      : "world"
+       #List of volumes where Ionization/scintillation are simulated (i.e. material = "LAr")
+       #Volumes used by the Hybrid Model
+       #Warning: make sure that names of LAr volumes are correct regarding the geometry used
+       volumeNames   : ["volTPCActive", "volCryostat", "volTPCPlaneVert","volXArapuca","volPMT","volFieldCage","volPDSstructure"]
+       #Corresponding stepLimits in mm for the volumes in the volumeNames list
+       stepLimits    : [0.3, 0.3, 0.3, 0.3, 0.3, 0.3, 0.3] 
+       gdmlFileName_ : "sbnd_v02_00_nowires.gdml"
```
ikatza, line 60: Is this necessary then?

If so it would be safer to do something like:
```
gdmlFileName_ : @geometry_or_something.gdmlFileName
```

ikatza:
Does this PR depend on the new, to be merged, geometry? If so I believe it would be good to mention that on the top comment of this PR (which can be edited).

sbndcode/JobConfigurations/standard/g4/OpHybrid_g4_refactored_sbnd.fcl
ggamezdiego: This line is something that will be possible/convenient to remove once the new geometry becomes the default one.

PR #111, opened 2021-06-09
HitDumper Updates

marcodeltutto:
This PR:
- Turns off SkipInd in the fast hit finder so that hits are created also for induction planes
- Adds settable fcl config to the HitDumper module that allows not saving the induction plane wire hits
- Switches to the optical hit finder in sbncode (PMT only)


mstancar (approved):
Looks good!

pgreen135:
trigger build

pgreen135:
CIs look good, I will include this in the next sbndcode release (next week).

PR #113, opened 2021-06-14
Removed deprecated T1053-version modules

wforeman:
There are some old versions of modules with "T1053"-based names. We now use the "SBND"-based versions of these modules, and the old T1053 versions haven't been developed in years (though continue to be maintained and built with cmake). This PR simply removes these deprecated modules.


wforeman:
Apologies for the revert -- accidentally pushed some unrelated changes to the wrong branch.

pgreen135:
trigger build

absolution1 (approved):
Nice one.
The hero we need but not the one we deserve.

PR #114, opened 2021-06-14
Restoring ability to toggle noise generation

wforeman:
Just a couple very minor updates. Restored the function of the "GenNoise" boolean in the simwire configuration, so users can turn on/off noise for tests. Also skipping the pedestal RMS routine if this is set to 0 in the fcl.

### DIFF
--- a/sbndcode/DetectorSim/detsimmodules_sbnd.fcl
+++ b/sbndcode/DetectorSim/detsimmodules_sbnd.fcl
@@ -7,25 +7,15 @@ sbnd_simwire:
  module_type:         "SimWireSBND"
  TrigModName:         "triggersim"
  DriftEModuleLabel:   "largeant"
- #NoiseFact:          0.15       #Noise Scale to use with histogram
- NoiseWidth:          62.4         #Exponential Noise width (kHz)
- NoiseRand:           0.1          #frac of randomness of noise freq-spec
- LowCutoff:           7.5          #Low frequency filter cutoff (kHz)
  CompressionType:     "none"       #could also be none		
+ BaselineRMS:         0.0         #ADC baseline fluctuation within channel        
  GenNoise:            true        # If false, NoiseService function is not called
- GetNoiseFromHisto:   false       # NO LONGER USED; set through NoiseService instead
- GenNoiseInTime:      true        # NO LONGER USED; set through NoiseService instead
 
  # the two settings below determine the ADC baseline for collection and induction plane, respectively;
  # here we read the settings from the pedestal service configuration,
  # so that we have only one maintenance point
  CollectionPed:       @local::sbnd_detpedestalservice.DetPedestalRetrievalAlg.DefaultCollMean # used to be 400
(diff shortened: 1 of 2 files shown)


mstancar:
The noise service already provides this option??  I have no objection if this is not possible through the service

wforeman:
Noise service options (DetectorSim/Services/noiseservices_sbnd.fcl) are rather complicated. Best I can tell, looks like they allow you to enable different individual types of noise like white noise, gaussian, coherent, etc. If I wanted to turn off the noise altogether, it's unclear what I should be changing in the noise service. So I think there's value in having a single module-level parameter that decides whether the noise service is called at all.

mstancar:
Agreed, just add comment to the fcl file then that specify that when this is set, the noise service is not called.

mstancar (approved):
Please also add a comment to the fcl file specifying that this parameter when set to false, tells the code not to call the noise service.

pgreen135:
trigger build

sbndcode/DetectorSim/detsimmodules_sbnd.fcl
```diff
@@ -12,9 +12,9 @@ sbnd_simwire:
  NoiseRand:           0.1          #frac of randomness of noise freq-spec
  LowCutoff:           7.5          #Low frequency filter cutoff (kHz)
  CompressionType:     "none"       #could also be none		
- GenNoise:            true        #gen noise...if false function not called
- GetNoiseFromHisto:   false
- GenNoiseInTime:      true
+ GenNoise:            true        # If false, NoiseService function is not called
+ GetNoiseFromHisto:   false       # NO LONGER USED; set through NoiseService instead
+ GenNoiseInTime:      true        # NO LONGER USED; set through NoiseService instead
```
absolution1, line 17: Could these be removed if they are no longer used?

wforeman: I was going to remove them, but didn't want to change too much at once. We could certainly remove "GenNoiseInTime" with no consequences, but "GetNoiseFromHisto" still appears in an uncommented code block where the noise spectrum histogram is retrieved... though this histogram is not used at all anywhere else.

absolution1: I'd vote for cutting them now.  They still exist in history so they can be reimplemented if needs be.

wforeman: Done!

pgreen135:
trigger build

pgreen135:
@wforeman looks like there's a c7 build error from an unused variable -- see: https://dbweb8.fnal.gov:8443/LarCI/app/ns:SBND/build_detail/phase_details?build_id=sbnd_ci/6518&platform=Linux%203.10.0-1160.21.1.el7.x86_64&phase=build&buildtype=slf7%20c7:prof

wforeman:
@pgreen135, fixed. The variable (fSampleRate) was also technically unused before as well, but the compiler counts a variable as "used" even if it appears in a commented-out block... weird!

pgreen135:
that is very strange! thanks for fixing it promptly

PR #115, opened 2021-06-15
HitDumper Updates

marcodeltutto:
This PR:
- Allows saving multiple ophit collections to the output tree
- Switches to saving both PMT and Arapucas ophits to the tree


pgreen135:
trigger build

PR #118, opened 2021-06-24
BackTracker messages are useless

ikatza:
We have all experience the barrage of BackTracker messages...

```
%MSG
%MSG-w BackTracker:  MCParticleHitMatching:gaushitTruthMatch@BeginModule  24-Jun-2021 18:18:27 CDT run: 1 subRun: 0 event: 50
caught exception
---- BackTracker BEGIN
  No sim::SimChannel corresponding to channel: 8658
---- BackTracker END
```

This commits ignores them

### DIFF
--- a/sbndcode/LArSoftConfigurations/messages_sbnd.fcl
+++ b/sbndcode/LArSoftConfigurations/messages_sbnd.fcl
@@ -169,6 +169,7 @@ sbnd_message_services_prod_debug:
 sbnd_message_services_prod_debug.destinations.LogDebugFile.categories:{
   default: { }
   GeometryBadInputPoint: { limit: 5 timespan: 1000}
+  BackTracker: { limit: 0 }
 }
 
 


ikatza:
Sadly, this doesn't change the ones occurring on cafmaker.

etyley:
If you want to fix the issue in the CAFMaker, the messaging service should be configured to use the sbnd one in the fcl (i.e. https://github.com/SBNSoftware/sbncode/blob/01540ccbee334b35226b07d93b15e9ffc834a070/sbncode/CAFMaker/cafmakerjob_sbnd.fcl#L34 should be changed to https://github.com/SBNSoftware/sbndcode/blob/d7028b95cec83b67cb87ab4c117d0f6db4556603/sbndcode/JobConfigurations/base/reco_sbnd.fcl#L63

The other services should probably also be changed to the sbnd defaults, but I guess that is another issue...

ikatza:
Hi @absolution1 do you agree with removing BackTracker messages?

pgreen135:
echoing the comments in the corresponding sbn PR, suppressing ALL messages from the back-tracker does not seem advisable even if these particular ones are not useful. I suggest closing this PR and opening an issue with LArSoft to address the underlying problem?

ikatza:
How about limiting them to 100 messages or similar?

The fact is they currently overwhelm all the terminal output, rendering other sensitive messages useless because they just get lost in the barrage of backtracker.

pgreen135:
That wouldn't resolve the problem though. I prefer the idea of following what Gianluca suggests in https://github.com/SBNSoftware/sbncode/pull/131 -- understanding the source, and potentially demoting this particular message to the debug stream.

ikatza:
You're correct, it doesn't solve the problem of the backtracker.

It's meant to solve the current problem of having useless output messages from everything else.

absolution1:
If I remember correctly, this 'error' springs up when dealing with energy depositions near the detector edges?  I faintly recall something along the lines of
 - edep happens in front of an edge channel
 - diffusion pushes some of the edep outside of the the edge channel
 - you get the complaint that the channel doesn't exist
 
I think I agree with @pgreen135 et. al. that this should be fixed at source rather than us blanket suppressing error messages from a crucial part of the analysis system.  
@ikatza I think it would be worth emailing scisoft, asking for clarification on how this can be solved (and possibly asking scisoft what other experiments have done in the past)

ikatza:
Fair enough. Let's close this.

PR #120, opened 2021-06-28
Fixing prodsingle T0

wforeman:
After doing some tests described below, it looks like prodsingle's T0 is relative to the beam gate time, so it should be set to 0.  Also, this parameter is in units of _nanoseconds_, not microseconds as I originally assumed. So by setting it to 1700 we were only offsetting the particles by 1.7 microseconds -- barely noticeable.

To test this, I simulated single 1.5 MeV electrons at the center of a TPC volume (X = -100cm) and looked at the times of reconstructed hits via `hit->PeakTime()`. For T0=0 and T0=1700, the result was the same, with hit times peaking at ~tick 1700, exactly in the center of our 3400-tick readout. But when setting T0=0.65e6 (0.65 ms, half a drift period), hit times shifted to tick 3000 as expected.


pgreen135:
trigger build

PR #121, opened 2021-07-01
PandoraShower Hit Inputs

etyley:
Since the input to pandora was changed to gaushit we can no longer use the standard set of cheating tools, we most define an SBND set with the appropriate overrides. I also moved the change to input label into the pandoramodules fcl rather than have it overridden in the reco fcl.


absolution1 (approved):
Nice one, @etyley 
I'll later on ask you to move all of the other overrides out of sbnd_reco.fcl ;)

pgreen135:
trigger build

PR #122, opened 2021-07-07
Adding 3drift legacy configuration for anyone wanting to process MC g…

wforeman:
Just adding `3drift_services_sbnd_legacy.fcl` for anyone who happens to be processing the old nu+cosmics MC generated with the previous drift length configuration. This file should be removed once a new MC production campaign takes place.

### DIFF
--- a/sbndcode/JobConfigurations/standard/gen/corsika/prodcorsika_proton_intime_filter.fcl
+++ b/sbndcode/JobConfigurations/standard/gen/corsika/prodcorsika_proton_intime_filter.fcl
@@ -26,5 +26,3 @@ physics.producers.larg4intime.InputLabels: [ "GenInTimeSorter:intime" ]
 #physics.producers.larg4intime.SparsifyTrajectory: "true"
 
 process_name: CosmicsCorsikaCMCGenAndG4InTime
-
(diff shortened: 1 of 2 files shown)


wforeman:
The second commit here makes the in-time cosmic generation file (prodcorsika_proton_intime_filter.fcl) use the default readout settings, instead of 3 drift windows, so the standard detsim/reco files will work over it. I _think_ this is what we want, since none of the other scripts use the 3-drift config without explicitly saying so in the filename ("*_3drift_windows.fcl"). Maybe @gputnam or @ikatza can confirm? 

I suggest making a second file with "_3drift_windows" appended to the name if we want to keep this option.

ikatza:
I confirm that with the last changes I can run `prodcorsika_proton_intime_filter_sce.fcl` and then run the `*_sce.fcl` all the way to reco2. 


As Will notes, this was previously failing with an error at detsim like this:

```
%MSG-s ArtException:  PostEndJob 07-Jul-2021 18:18:37 CDT ModuleEndJob
---- DetectorClocksServiceStandard BEGIN
  Found historical value disagreement for G4RefTime ... -1700 != -2800
---- DetectorClocksServiceStandard END
%MSG
```

PR #123, opened 2021-07-08
Fix to xarapuca dimensions swap on the geometry

ikatza:
This issue was found by @rodralva, he reached out to me for guidance. I'm pushing this fix with the hope that gets merged ASAP because I need it for an upcoming production. 

The new geometry mixes up the `x` and `z` coordinates of the XARAPUCAS. Thus, the dimensions of the XARAPUCAS relevant to the solid angle computation become:
```
opDet.Length(): 0.2, opDet.Height(): 20.6
```

The consequence is a sharp drop off of the collected simphotons, again, @rodralva owns the credit of finding this issue.

Talking about the active window dimensions:
- The **width**, which is `2.0 mm` runs along the     `x` axis.
- The **height**, which is      `206.0 mm` runs along the `y` axis.
- The **length**, which is      `76.0 mm` runs along the     `z` axis. 

In here **width, height, length** `(x,yz)` refer to the ones understood by larsoft. 

The geometry uses **width, length, thickness**, where **width** refers to the short side and **length** to the long side. I personally prefer larsoft's usage, as I find them easier to understand and remember.

This PR doesn't change the naming convention used on the geometry, only flips the right stuff to get simulations OK again.

```
opDet.Length(): 7.6, opDet.Height(): 20.6
```

### DIFF
--- a/sbndcode/Geometry/gdml/sbnd_v02_00.gdml
+++ b/sbndcode/Geometry/gdml/sbnd_v02_00.gdml
@@ -12750,10 +12750,10 @@
                 </volume>
         </structure>
 	<solids>
-		<box name="XArapucaOut" lunit="mm" x="96" y="240" z="20"/>
-		<box name="XArapucaFilter" lunit="mm" x="76" y="206" z="2"/>
-		<box name="XArapucaHole" lunit="mm" x="76" y="206" z="3.0009999999999999"/>
-		<box name="XArapucaCover" lunit="mm" x="76" y="8" z="1"/>
+                <box name="XArapucaOut" lunit="mm" x="20" y="240" z="96"/>
+                <box name="XArapucaFilter" lunit="mm" x="2" y="206" z="76"/>
+                <box name="XArapucaHole" lunit="mm" x="3.0009999999999999" y="206" z="76"/>
+                <box name="XArapucaCover" lunit="mm" x="1" y="8" z="76"/>
 		<subtraction name="XArapucaExternalWalls">
 			<first ref="XArapucaOut"/>
 			<second ref="XArapucaHole"/>
--- a/sbndcode/Geometry/gdml/sbnd_v02_00_base.gdml
+++ b/sbndcode/Geometry/gdml/sbnd_v02_00_base.gdml
@@ -4563,10 +4563,10 @@
 
 	<solids>
 
-		<box name="XArapucaOut" lunit="mm" x="XArapucaOut_w" y="XArapucaOut_l" z="XArapucaOut_t"/>
-		<box name="XArapucaFilter" lunit="mm" x="XArapucaFilter_w" y="XArapucaFilter_l" z="XArapucaFilter_t"/>
-		<box name="XArapucaHole" lunit="mm" x="XArapucaFilter_w" y="XArapucaFilter_l" z="XArapucaFilter_t+XArapucaFilterDepth+_eD_"/>
-		<box name="XArapucaCover" lunit="mm" x="XArapucaFilter_w" y="distanceBetweenXArapucaFilters" z="XArapucaFilterDepth"/>
+                <box name="XArapucaOut" lunit="mm" x="XArapucaOut_t" y="XArapucaOut_l" z="XArapucaOut_w"/>
+                <box name="XArapucaFilter" lunit="mm" x="XArapucaFilter_t" y="XArapucaFilter_l" z="XArapucaFilter_w"/>
+                <box name="XArapucaHole" lunit="mm" x="XArapucaFilter_t+XArapucaFilterDepth+_eD_" y="XArapucaFilter_l" z="XArapucaFilter_w"/>
+                <box name="XArapucaCover" lunit="mm" x="XArapucaFilterDepth" y="distanceBetweenXArapucaFilters" z="XArapucaFilter_w"/>
 
 		<subtraction name="XArapucaExternalWalls">
 			<first ref="XArapucaOut"/>
--- a/sbndcode/Geometry/gdml/sbnd_v02_00_nowires.gdml
+++ b/sbndcode/Geometry/gdml/sbnd_v02_00_nowires.gdml
@@ -12750,10 +12750,10 @@
                 </volume>
         </structure>
 	<solids>
-		<box name="XArapucaOut" lunit="mm" x="96" y="240" z="20"/>
-		<box name="XArapucaFilter" lunit="mm" x="76" y="206" z="2"/>
-		<box name="XArapucaHole" lunit="mm" x="76" y="206" z="3.0009999999999999"/>
-		<box name="XArapucaCover" lunit="mm" x="76" y="8" z="1"/>
+                <box name="XArapucaOut" lunit="mm" x="20" y="240" z="96"/>
+                <box name="XArapucaFilter" lunit="mm" x="2" y="206" z="76"/>
+                <box name="XArapucaHole" lunit="mm" x="3.0009999999999999" y="206" z="76"/>
+                <box name="XArapucaCover" lunit="mm" x="1" y="8" z="76"/>
 		<subtraction name="XArapucaExternalWalls">
 			<first ref="XArapucaOut"/>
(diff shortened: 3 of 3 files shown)


ikatza:
Just to be clear, first commit has manual changes to the base file.

The second has the output of:
```
 preparseGDML sbnd_v02_00_base.gdml -w
```

I didn't manually updated myself.

pgreen135:
@ikatza is this not essentially rotating the optical detectors in the geometry by 90 degrees? were they facing the wrong way before? or are they now facing the wrong way?

ikatza:
It's not. The length is wrong:

```
opDet.Length(): 0.2
```

```
opDet.Length(): 7.6,
```

as I report above.

pgreen135:
right, because the length function gets the z coordinate; but isn't swapping the x <-> z size is changing the direction the arapucas face relative to everything else in the detector? if you draw the gdml

ikatza:
Oh you mean for the purposes of drawing the geometry, then yes. I believe you're right.

I don't know how to draw the geometry, but I expect this change to fix the drawing too.

pgreen135:
> Oh you mean for the purposes of drawing the geometry, then yes. I believe you're right.
> 
> I don't know how to draw the geometry, but I expect this change to fix the drawing too.

yes, exactly -- the physical orientation of the sensitive window in the gdml is important to also have correct, as that impacts the full light simulation if updated parameterisations / libraries are needed

I will let @gustavogx @marcodeltutto comment on this change

marcodeltutto:
Thanks @ikatza and @rodralva for finding the problem!
@pgreen135 is correct that this PR would rotate the arapucas in the geometry, making them face the wrong way.

From current `develop`:
Screen Shot 2021-07-08 at 10 23 06 AM

From this PR:
Screen Shot 2021-07-08 at 10 25 12 AM

What happens in geometry `v02_00` compared to `v01_05` is that each PDS module (8 xarapucas and 5 pmts) are constructed together in the gdml, and then rotated and placed in position. So the X and Z dimensions in the gdml are correct in develop, because that are then rotated. This was not happening in `v01_05`. The problem now is that the `Length` method in larsoft is returning not what we expect.

pgreen135:
@marcodeltutto can these not be constructed more analogously to v01_05? it is concerning that getting dimensions etc. from the geometry service are returning unexpected things

marcodeltutto:
I will let @gustavogx comment at this point, as I haven't worked on this myself.

ikatza:
In favour of larsoft, the geometry call is consistent with what I've outline above (after making an edit).

The geometry function `Length()` in larsoft:
https://nusoft.fnal.gov/larsoft/doxsvn/html/classgeo_1_1OpDetGeo.html#a99581ce607905340d1b3a83f236a0c60

which calls `HalfL`
https://nusoft.fnal.gov/larsoft/doxsvn/html/classgeo_1_1OpDetGeo.html#aa39d05069f035101cd464303d3fbac54
```cpp
   {
      TGeoBBox const* pBox = asBox();
      return pBox? pBox->GetDZ(): 0.0;
    }
```

defined as `dZ`. `Width()` relates to `X`, and `Height()` to `Z`.

rodralva:
I just ran 1k muon events with Iker's updates, they seem to fix the problem. This is the raw simphotons arriving at the XArapuca and PMT channels from the G4 stage:

image
image

gustavogx:
Joining the discussion a little late, but here we go...

The Arapuca's orientation are correct.

The problem is that the whole PDS box is designed in XY plane and then rotated 90 deg along the Y axis either clockwise or anti-clockwise, when positioning them in the APAs. Look for line 5793 in the _base file (or search for volPDSstructure ).

The optical library is assuming that the dimensions you see in the viewer correspond to the objects own x y z, but that is usually not the case. Many objects are placed using rotations.Moreover, I'm pretty sure this is affecting the PMT's as well. The only reason no one noticed is because they are hemispheres.

I'm afraid there's no quick fix here. If I just swap the coordinates as you mentioned, there will be solid clashes.

The proper fix is to go over every piece of geometry in the PDS box and swap x and z in all of its elements, so that it ends up facing the X direction, then use only one 180deg rotation in the Y direction to place it on the other side.

I will talk with Marina.

gustavogx:
> I just ran 1k muon events with Iker's updates, they seem to fix the problem. This is the raw simphotons arriving at the XArapuca and PMT channels from the G4 stage:
> 
> image
> image

You see, since the whole PDS box is rotated, the PMTs are also affected. It is hard so see because of the log scale, but it is there

gustavogx:
> Thanks @ikatza and @rodralva for finding the problem!
> @pgreen135 is correct that this PR would rotate the arapucas in the geometry, making them face the wrong way.
> 
> From current `develop`:
> Screen Shot 2021-07-08 at 10 23 06 AM
> 
> From this PR:
> Screen Shot 2021-07-08 at 10 25 12 AM
> 
> What happens in geometry `v02_00` compared to `v01_05` is that each PDS module (8 xarapucas and 5 pmts) are constructed together in the gdml, and then rotated and placed in position. So the X and Z dimensions in the gdml are correct in develop, because they are then rotated. This was not happening in `v01_05`. The problem now is that the `Length` method in larsoft is returning not what we expect.

That would not be a first. We actually avoid using rotations precisely because many of the LArSoft's also do (avoid it). In this case, I thought I would be ok (and I was wrong).

gustavogx:
I'm working on it

pgreen135:
> The proper fix is to go over every piece of geometry in the PDS box and swap x and z in all of its elements, so that it ends up facing the X direction, then use only one 180deg rotation in the Y direction to place it on the other side.

Thanks for looking into this @gustavogx @mrguzzo. This approach sounds the most appropriate to preserve the directions as the geometry service, and hence the semi-analytic light simulation etc., expects.

ikatza:
The discrepancies on the observed photons for the PMTs on the plots Rodrigo showed are not related to the rotated boxes. The plots he made were created with the semi-analytic method. I should have stresses this, the `opDet.Length()` and `opDet.Heigth()` I report are those used on `larsim` for the solid angle computations.

The radius of the dome for the PMTs is taken from a fcl file, which hasn't changed. As far as I can tell their positions are OK. 

It seems to me that, overall there are less detected photons on the new geometry. I believe one of the reasons behind this is the increase in drift distance on the latest geometry, and the fact that foils moved a couple of cm away from the optical detectors.

gustavogx:
Hi guys, just reporting that I will push the correction to my feature branch in a few minutes. 

Meanwhile, I will point out that this issue could be worked out with the LArSoft group. There are some limitations that we have to give in when designing the geometry, which are coming from LArSoft. Namely, 1) rotations are mostly accounted for in the volume's  own node, but it should be accumulated from all nodes above it, in all cases, and 2) reflections are not accounted for at all. If there is a good reason in LArSoft for not doing this, I would like to know cause it shifts a heavy burden to the gdml file.

I am willing to talk with them and even collaborate with some fixes, I just don't know with whom I should start. Could someone point me in the right direction?

pgreen135:
> Hi guys, just reporting that I will push the correction to my feature branch in a few minutes.
> 
> Meanwhile, I will point out that this issue could be worked out with the LArSoft group. There are some limitations that we have to give in when designing the geometry, which are coming from LArSoft. Namely, 1) rotations are mostly accounted for in the volume's own node, but it should be accumulated from all nodes above it, in all cases, and 2) reflections are not accounted for at all. If there is a good reason in LArSoft for not doing this, I would like to know cause it shifts a heavy burden to the gdml file.
> 
> I am willing to talk with them and even collaborate with some fixes, I just don't know with whom I should start. Could someone point me in the right direction?

I suggest opening an issue here: https://cdcvs.fnal.gov/redmine/projects/larsoft/issues, and it will be directed towards the right people.

pgreen135:
As there is now a new PR addressing this issue (https://github.com/SBNSoftware/sbndcode/pull/124) I am going to close this PR

gustavogx:
I just updated feature/NewGeometry_v2 branch with my correction. Nothing else has changed. I already made a pull request for it.

PR #124, opened 2021-07-09
Reworked PDS boxes and X-ARAPUCAS

gustavogx:
Reworked PDS boxes and X-ARAPUCAS to use rotations in a way to NOT confuse the Light Simulation

### DIFF
--- a/sbndcode/Geometry/gdml/sbnd_v02_00.gdml
+++ b/sbndcode/Geometry/gdml/sbnd_v02_00.gdml
@@ -12750,16 +12750,16 @@
                 </volume>
         </structure>
 	<solids>
-		<box name="XArapucaOut" lunit="mm" x="96" y="240" z="20"/>
-		<box name="XArapucaFilter" lunit="mm" x="76" y="206" z="2"/>
-		<box name="XArapucaHole" lunit="mm" x="76" y="206" z="3.0009999999999999"/>
-		<box name="XArapucaCover" lunit="mm" x="76" y="8" z="1"/>
+		<box name="XArapucaOut" lunit="mm" x="20" y="240" z="96"/>
+		<box name="XArapucaFilter" lunit="mm" x="2" y="206" z="76"/>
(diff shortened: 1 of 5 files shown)


pgreen135:
@ikatza @rodralva could you repeat your test with the muons to double-check that this resolves the issue?

gustavogx:
A comment: I used the information provided by @ikatza in the other issue (#123):

"The radius of the dome for the PMTs is taken from a fcl file, which hasn't changed. As far as I can tell their positions are OK."

And applied rotations to the PMT only. This is unavoidable since cylinder solids can only be generated with its symmetry axis in the Z direction. But, since Iker clarified that the PMTs info is taken from the fcl file, it shouldn't be a problem.

ikatza:
Hi, these changes don't work. When printing the XARAPUCA dimensions I get:

```
Length:0.2, Height:20.6
```

I suppose larsoft expects to get an `x, y, z` vector in that order. And the latest changes swapped `z` to be in the first entry, and `x` to be on the last.

gustavogx:
I don't believe it. The coordinates should be read by the parser in an
order-independent fashion. What is going on here?

I will change the values instead. Give me some time.
G.

Em sex., 9 de jul. de 2021 às 13:06, Iker de Icaza ***@***.***>
escreveu:

gustavogx:
I just pushed a fix. Could you try it out @ikatza  ?

ikatza:
Yes! 

Fixed:
```
Length:7.6, Height:20.6
```

pgreen135:
There will be a new sbndcode tag on Monday, I will include this. Thanks everyone for finding and fixing this so promptly!

rodralva:
Everything seems fine
image
It's very preliminary, but it seems that there's a bit less light arriving at detectors with the new geometry (altough this was just a small 1k event sample). [This is the fcl cofiguration](https://github.com/SBNSoftware/sbndcode/blob/develop/sbndcode/JobConfigurations/standard/gen/genie/other_flux_config/flux_configf/prodgenie_nu_singleinteraction_tpc_gsimple-configf-v1.fcl) for the muons I've been using.

PR #128, opened 2021-07-26
Override larsoft and assign view using plane number for signal shape sim

wforeman:
LArSoft is confusing the "View" (U/V/Y) for planes 0/1 in TPC 0 in the new geometry, resulting in the simulated signal shape being swapped for these two planes. The result is a ~10-fold increase in noise hits in tpc0/plane0 due to mis-matched signal shapes in the de-convolution stage. This fix simply overrides LArSoft's strange way of assigning the View, and instead explicitly uses the plane numbers (0=U, 1=V).


pgreen135:
trigger build

pgreen135:
now we see CI test differences as expected! @wforeman I presume these make sense? (e20, ignore the c7)

pgreen135 (approved):

PR #129, opened 2021-07-28
Simple-Flash Match: multi-matcher, unbiased, tuneable, fast.

ikatza:
Depends on https://github.com/SBNSoftware/sbncode/pull/152
Where discussion should happen.

I've asked @pgreen135 to update `sbnd_data`

### DIFF
--- a/sbndcode/FlashMatch/flashmatch_sbnd.fcl
+++ b/sbndcode/FlashMatch/flashmatch_sbnd.fcl
@@ -11,7 +11,7 @@ sbnd_simple_flashmatch:{
   FlashStart: -0.02 # us, wrt flash time
   FlashEnd: 0.08 # us, wrt flash time
   SelectNeutrino: true
-  OnlyPrimaries: true
+  OnlyCollectionWires: false
   UseUncoatedPMT: true
   UseARAPUCAS: true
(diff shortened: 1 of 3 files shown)


sbndcode/FlashMatch/flashmatch_sbnd.fcl
```diff
@@ -39,21 +40,28 @@ sbnd_simple_flashmatch:{
   score_hist_low: 0.
   score_hist_up: 50.
 
-  n_bins: 40
-  DriftDistance: 202.05
+  DriftDistance: 202.05 # only used for metrics creation
```
pgreen135, line 43: as the new geometry now gives the correct drift distance, this could be done using the geometry service

ikatza: That's the value I get from the geometry. I use whatever the geometry outputs for normal running, and don't use the fcl parameter. 

But I need to define this on a fcl to have it on the Python script that creates the metrics file.

pgreen135: got it, thanks for the clarification

pgreen135 (approved):

PR #133, opened 2021-08-04
XArapucas Updates

rodralva:
XArapuca Updates in the OpDetSim Module related to issues [#78](https://github.com/SBNSoftware/sbndcode/issues/78), [#80](https://github.com/SBNSoftware/sbndcode/issues/80) :

-Added new sampling field to the sbnd_pds_mapping.json file.
-Modified digitizer, op hit finder, wvf analyzer and trigger modules to differentiate daphne xarapuca channels with their own frequency. 
-Updated test bench XArapuca single electron response.

The data driven response is loaded from sbnd_data  ([OpDetSim/digi_arapuca_sbnd.root](https://github.com/SBNSoftware/sbndcode/blob/feature/rodrigoa_XArapucasUpdates/sbndcode/OpDetSim/digi_arapuca_sbnd.fcl#L27)). The file needs to be updated adding the vectors with the single electron response data before the CI tests.

PR #138, opened 2021-08-09
Do not use xarapucas for fm

ikatza:
This is to address a potential issue that may arise from the change of the sampling frequency of (most of) the XARAPUCAS in: https://github.com/SBNSoftware/sbndcode/pull/133

As such, it should only be merged after that one.

The time resolution is substantially different, from 2 ns to 12.5 ns. One of the outcomes is that XARAPUCAS could potentially shift the peak towards later times. Surely there are other potential issues.

I've tested running without using xarapucas, the change of this PR, on 10k cosmics events and the FM still works OK. A more elaborate solution that allows us to mix different time resolutions needs to be worked out.

### DIFF
--- a/sbndcode/FlashMatch/flashmatch_sbnd.fcl
+++ b/sbndcode/FlashMatch/flashmatch_sbnd.fcl
@@ -14,7 +14,7 @@ sbnd_simple_flashmatch:{
   OnlyCollectionWires: false
   ForceConcurrence: true
   UseUncoatedPMT: true
-  UseARAPUCAS: true
+  UseARAPUCAS: false
   InputFileName: "FlashMatch/fm_metrics_sbnd.root"
   NoAvailableMetrics: false
   MakeTree: false


pgreen135:
I will merge this once the corresponding xarapuca changes are merged. The x-arapucas certainly ought to be incorporated in the flash matching --- is this simply an issue of retraining the algorithm, or are there more fundamental problems with this module? 

Also, could you open an issue for this? so we can keep track it

wesketchum:
I've attached it to the current XARAPUCA issue (though that may not be the best one). Feel free to unlink and assign better.

ikatza:
I don't think retraining will be necessary. 

Currently the module assumes that all ophits have the same time resolution, so that needs to be changed somehow. I don't think it will be too complicated to fix and maybe it won't be much of a problem. I just don't have time to work on this now.

Switching to only using PMT is an acceptable workaround, but of course the intention is to incorporate XARAPUCAS. Linking to that issue is OK for now, I'll get back to this in the future.

pgreen135 (approved):

PR #155, opened 2021-08-18
Update PDFastSim_sbnd.fcl

ggamezdiego:
Updating the triplet-time to 1300ns in the newlarg4.
Notice that scintillation decay times are (re)defined in a new place "scinitllationtime_tool.fcl". The reason for this redefinition (instead of keep using the "larproperties.fcl" values) is that the logic of the new LArG4  is to try to minimize the dependence on services, since they interfere with multi-threading and make it harder to understand how the code is behaving. The new philosophy on any new code that goes in is to move configuration parameters to sit in the place where they are used.


jzennamo (approved):

jzennamo:
Looks good, thanks!

PR #134, opened 2021-08-04
Migration to the Refactored LArG4

marcodeltutto:
This PR contains all the necessary fcls and code to migrate to the new LArG4. 
Changes are presented on docdb [22944](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=22944).

Note that with this PR, the `standard_g4_sbnd.fcl` will run the new LArG4. To run the legacy larg4, fcl `legacy_g4_sbnd.fcl` should be used. All the `g4_*.fcl` files run the new LArG4. To run the legacy LArG4, replace `g4_*.fcl` with `legacy_g4_*.fcl`.
All the legacy g4 fcl files are in `JobConfigurations/standard/g4/legacy/*`

A new folder called `LArG4` is added. This contains all the g4-related fcl configuration, as well as the physics list to be used in `LArG4/PhysicsLists/`. In this `LArG4` folder:
- `larg4_sbnd.fcl`: Configuration of the LArG4 module
- `larg4_services_sbnd.fcl`: Configurations needed for the new LArG4 
- `ionandscint_sbnd.fcl`: Configuration for the ionization and scintillation part
- `simdrift_sbnd.fcl`: Configuration for the electron drifting module
- `mcreco_sbnd.fcl`: Configuration for MC reconstruction
- `PhysicsListLoader`: a dummy module to load the physics list, needed before running LArG4

To run with the optical hybrid mode, fcl file `g4_ophybrid_sbnd.fcl` can be used in place of `standard_g4_sbnd.fcl`.

### Open Items

- [x] Use **correlated** light model? -> Yes, as PR #127 was merged.
- [x] Use **optical hybrid** mode? -> Added to default g4 workflow.
- [x] ~~**CRT** part. -> Waiting for PR to be opened.~~

### FHICL files
Suggested reviewer: @jzennamo 
Standard fcls now run the new larg4. All the legacy fcls are in `sbndcode/JobConfigurations/standard/g4/legacy/`, and similar for `detsim`.

### Optical
Suggested reviewer: @ggamezdiego  
- `sbndcode/JobConfigurations/standard/standard_g4_sbnd.fcl`: Runs in optical hybrid mode by default.
- `sbndcode/LArSoftConfigurations/photpropservices_sbnd.fcl`: Adds optical library for light outside the active volume.
- `sbndcode/LArSoftConfigurations/PDFastSim_sbnd.fcl`: Configuration for semi-analytical and optical library. 
[...]

PR #156, opened 2021-08-18
Use new module label for sim channels

henrylay97:
Anatree standard configuration requires these changes to run on files produced using sbndcode `v09_28_01` as the module label for SimChannels was changed by PR #134


pgreen135 (approved):
Thanks Henry for fixing this

PR #162, opened 2021-08-19
Feature/jzennamo recofix

jzennamo:
Adding two fixes for getting reco to work for SBND:
1. Removed opT0Finder which is currently seg faulting
2. Change the SimChannel label to match the new LArG4 

These fixes enabled me to run the full workflow through to reco2


wesketchum (commented):
Look good to me.

wesketchum (approved):

PR #164, opened 2021-08-19
Capping the number of Flux Files Used Due to Stash Cache Latency Issues

jzennamo:
Due to the latency of stash cache, the default number of flux files being used (~150) is causing long delays in the configuration of GENIE jobs. This limits the number of flux files to be a bit more than ~5. This enables us to run GENIE generation jobs in <1-2 hrs.


wesketchum (approved):

mroda88 (approved):
looks good to me. Any idea about how we could check if this has an impact on the physics without a full generation? Or am I simply overthinking the problem and there is no chance for this to happen?

PR #170, opened 2021-09-01
Match resource ranges between quick and seq tests

henrylay97:
We regularly update the resource usage ranges for the quick (automatic trigger) tests. The equivalents for the sequential tests are very out-of-date so I've updated them to match the same as the quick tests.


pgreen135 (approved):
Makes sense, thanks Henry

PR #172, opened 2021-09-01
Create fcl_checks test for the ci

henrylay97:
Have created a ci_test that performs a standard set of checks on fcl files:
- Firstly, it checks that the fcl can be parsed (when fully expanded)
- Secondly, it compares the output of a `fhicl-dump` to a reference version

The list of files to check is provided in `sbndcode/test/ci/fcl_file_checks.list`. The current list is arbitrary and created by myself, this is why the PR is in draft mode. I will liase with production and others offline to produce a better list of files we want to test and push this in a separate commit.

The update for the reference file is different to the other regression tests and I will let @vitodb give his opinion on that. The reason for not using the standard method is the test doesn't fit the template of the other regression tests (in which one stage of the sim/reco chain is performed). My approach defines a separate test in `ci_tests.cfg` which uses the same script with an `--update-ref-files` flag. This test is then included in the `generate_reference_sbndcode` suite.

The indentation in the `fcl_checks.sh` script looks incorrect on github but is fine when opened offline, not sure why.


cjbackhouse:
> The indentation in the `fcl_checks.sh` script looks incorrect on github but is fine when opened offline, not sure why.

Probably you have a mix of spaces and tabs. Prefer all-spaces.

henrylay97:
Thanks Chris, have pushed a correction :)

henrylay97:
I have updated the list of fcls to based on the production spreadsheet. This can be further updated as we go along (I will ask for input from others) but this seems like a reasonable list to begin with. Can now be reviewed, @vitodb, but there is no immediate rush! I imagine you'll have some comments on my reference file updating method ;)

henrylay97:
@vitodb I've updated this to use a tar for the outputs. I've tested this works, do you think we should hold off to try and amend the way the references are handled or merge as is and work that in later on?

vitodb:
> @vitodb I've updated this to use a tar for the outputs, sorry it took me so long. I've tested this works, do you think we should hold off to try and amend the way the references are handled or merge as is and work that in later on?

@henrylay97 I'm working on some updates to allow this new test to use the CI mechanism to handle the reference file.
I'm positive to get this done shortly.
If this feature can wait a bit longer we could merge the full featured version.

henrylay97:
Absolutely, I think that's a good plan, thanks as ever!

vitodb:
With the last update I made this PR now is able to use the CI mechanism to handle reference files for `fcl_checks` test.
A test CI build using the code provided by Henry to update references for `fcl_checks` test and running this test is available in this [CI dashboard](https://dbweb9.fnal.gov:8443/TestCI/app/ns:sbnd/view_builds/index?offset=0&builds=sbnd_ci%2F7703&builds=)
CI build `sbnd_ci/7702` has been used to test `fcl_checks` reference update,
CI build `sbnd_ci/7703` has been used to test `fcl_checks` regression test.
 
I'm approving this PR, but I'll be glad if Henry can have a final check on my updates before getting this PR merged.

vitodb (approved):

henrylay97:
Looks great to me, thanks for your help with the updates Vito! All ready to be merged from my point of view. 

@wesketchum @jzennamo @chilge (and Mateus who I can't find on git). Would be great to have some further input on which fcl files you would like tested. Clearly this can be a rolling thing but would be good to have the fcls you want for MCP2021C in place :) The current list is defined in [sbndcode/ci/fcl_file_checks.list](https://github.com/SBNSoftware/sbndcode/blob/6ddee7d992b72ef79e6d57fe422a3ccc3a9c1221/test/ci/fcl_file_checks.list).

chilge:
@rennney is this something that can be easily ported over to `icaruscode`?

jzennamo:
Re:fcl_file_checks.list I would propose the following list, these inherit the others but also configure the SCE service so are more complete (and what we run in production) :
```
- g4_sce.fcl
- detsim_sce.fcl
- reco1_sce.fcl
- reco2_sce.fcl
- cafmakerjob_sbnd_sce_geniewgt.fcl
```

henrylay97:
Thanks @jzennamo I have implemented your suggestion and I think the gen fcls cover all the production routes too, (note these can be updated _very_ easily whenever we need). This is ready to merge :)

PR #179, opened 2021-09-10
Feature/lynnt comm

lynnt-uchicago:
pull request for the crossing muon modules for comissioning. I also have a readme included for my feature branch, but not sure if there's a way to exclude it from the pull request?


mstancar (approved):
Looks good

sbndcode/Commissioning/MuonTrackProducer_module.cc
```diff
@@ -0,0 +1,709 @@
...
+#include <algorithm>
+#include <stdlib.h>
+#include <math.h>
+#include <vector>
+#include <iostream>
+#include <cmath>
+#include <bitset>
+#include <memory>
+
+#define PI 3.14159265
```
marcodeltutto, line 44: I'd suggest we don't re-define `PI`. You can remove this line and use `M_PI` in your code.

sbndcode/Commissioning/MuonTrackProducer_module.cc
```diff
@@ -0,0 +1,709 @@
...
+
+   // parameters from the fcl file 
+   std::string fHitsModuleLabel; 
+   int fHoughThreshold;
+   int fHoughMaxGap;
+   int fHoughRange;
+   int fHoughMinLength;
+   int fHoughMuonLength;
+   int fEndpointRange; 
+   std::vector<int> fKeepMuonTypes = {0, 1, 2, 3, 4, 5};
```
marcodeltutto, line 117: Can you please add a comment next to these variables? This will make it easier for people to understand what they and how they are used later. Also describing the muon types, or referring to another part of the code where they are defined.

sbndcode/Commissioning/MuonTrackProducer_module.cc
```diff
@@ -0,0 +1,709 @@
...
+      art::fill_ptr_vector(hitlist, hitListHandle);
+      nhits = hitlist.size();
+   }
+   else {
+      std::cout << "Failed to get recob::Hit data product." << std::endl;
+      nhits = 0;
+   }
+   
+   // obtain collection hits, perform hough transform, obtain tracks, and save col track info 
+   ResetCollectionHitVectors(20);
```
marcodeltutto, line 174: Do we want this `20` to be hardcoded, or also configurable via a fcl parameter? (also later in the code)

sbndcode/Commissioning/MuonTrackProducer_module.cc
```diff
@@ -0,0 +1,709 @@
...
+                     (outhit_idx.at(j)).insert( (outhit_idx.at(j)).end(),  (outhit_idx.at(i)).begin(),  (outhit_idx.at(i)).end()); 
+                     (outhit_idx.at(i)).clear();
+                  }
+               } 
+            }
+         }
+      } // end of j loop 
+   } // end of i loop 
+
+   for (int i=0; i<int(outlines.size()); i++){
```
marcodeltutto, line 453: I'd suggest we don't use C-casting (with `int()`). In this case, you can  use `size_t`: `for (size_t  i = 0; i < outlines.size(); i++) {`

sbndcode/Commissioning/MuonTrackProducer_module.cc
```diff
@@ -0,0 +1,709 @@
...
+
+void MuonTrackProducer::ResetCollectionHitVectors(int n) {
+   hit_02.clear(); 
+   hit_12.clear(); 
+   lines_02.clear(); 
+   lines_12.clear(); 
+   hit_idx_02.clear();
+   hit_idx_12.clear();
+
+   hit_02.reserve(3000);
```
marcodeltutto, line 486: Just wanted to make sure that having these numbers hardcoded is intentional? Also in `ResetInductionHitVectors`

README.md
```diff
@@ -0,0 +1,28 @@
+# Crossing muon tracks
```
marcodeltutto, line 1: Having this README file would be nice, why don't you move it to the `sbndcode/Commissioning` directory?

marcodeltutto (changes requested):
Thank you @lynnt-uchicago! Looks great! I added a few minor comments if you don't mind taking a look at them. As per the README file, I'd suggest you move it to the `sbndcode/Commissioning` directory instead of removing it, it's nice to have documentation written down!

marcodeltutto:
@pgreen135 Just a heads up that this PR depends on SBNSoftware/sbnobj#26

pgreen135:
thanks @marcodeltutto, I'll make sure these go in together

sbndcode/Commissioning/MuonTrackProducer_module.cc
lynnt-uchicago: Question about these actually: these hard coded numbers are an estimate of how many induction/collection hits there are in a given event. I could easily move them to a fcl parameter (like I did with another hardcoded number), but unsure if that's actually more helpful in the long run?

lynnt-uchicago:
realized that I need to add the muonhit info from the association in the producer, will be working on that!

sbndcode/Commissioning/MuonTrackProducer_module.cc
cjbackhouse: Butting in: I don't think anyone is going to be interested in configuring `reserve()` calls from fcl. It's solely a performance optimization, which probably has negligible effect here. I would just leave them, or potentially remove the `reserve()` calls since I doubt they really matter.

marcodeltutto: Yes, no need to make this fhicl configurable. I think there is some benefit in having these `reserve` as the vectors are filled with `push_back` so having the memory reserved beforehand will save some time. I would just leave them there too as they are. Thanks!

.gitignore
```diff
@@ -58,3 +58,6 @@ Thumbs.db
 *.bak
 *.backup
 RepositoryTag
+sbndcode/Commissioning/safe/ACFilter_module.cc
+sbndcode/Commissioning/safe/ACProducer_module.cc
+sbndcode/Commissioning/safe/run_acproducer.fcl
```
marcodeltutto, line 63: I think these lines are for you setup @lynnt-uchicago, can we remove them from this PR so they don't end up in our common repo? Thanks!

marcodeltutto (approved):
Looks good! Thank you! Please just remove the three lines in the `.gitignore` file, otherwise this is good to go!

pgreen135:
trigger build SBNSoftware/sbnobj#26

PR #180, opened 2021-09-14
Adding tool for PMT gain fluctuations

fjnicolas:
This PR adds an art tool to simulate PMT gain fluctuations (algorithm taken from icaruscode/PMT/Algorithms/PMTsimulationAlg.h). More details can be found in [docdb-23286](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=23286). Code modifications:
- Added [PMTAlg](https://github.com/SBNSoftware/sbndcode/tree/feature/fnicolas_sprfluc/sbndcode/OpDetSim/PMTAlg) directory containing the art tool for gain fluctuations
- Added modified version for the 'AddSPE' function in [DigiPMTSBNDAlg.cc](https://github.com/SBNSoftware/sbndcode/blob/feature/fnicolas_sprfluc/sbndcode/OpDetSim/DigiArapucaSBNDAlg.cc#L261)

Note that the MakeGainFluctuations parameter in [digi_pmt_sbnd.root](https://github.com/SBNSoftware/sbndcode/blob/feature/fnicolas_sprfluc/sbndcode/OpDetSim/digi_pmt_sbnd.fcl#L36) is set to true, so **SER fluctuation are simulated by default with this PR**.

### DIFF
--- a/sbndcode/OpDetSim/DigiPMTSBNDAlg.cc
+++ b/sbndcode/OpDetSim/DigiPMTSBNDAlg.cc
@@ -61,7 +61,7 @@ namespace opdet {
     }
 
     if(fParams.MakeGainFluctuations){
-      fPMTGainFluctuationsPtr = art::make_tool<opdet::PMTGainFluctuations>(fParams.GainFluctuationsPset);
+      fPMTGainFluctuationsPtr = art::make_tool<opdet::PMTGainFluctuations>(fParams.GainFluctuationsParams);
     }
 
     saturation = fParams.PMTBaseline + fParams.PMTSaturation * fParams.PMTChargeToADC * fParams.PMTMeanAmplitude;
@@ -145,10 +145,7 @@ namespace opdet {
         tphoton = ttsTime + simphotons[i].Time - t_min + ttpb + fParams.CableTime;
         if(tphoton < 0.) continue; // discard if it didn't made it to the acquisition
         timeBin = std::floor(tphoton*fSampling);
-        if(timeBin < wave.size()){
-          if(fParams.MakeGainFluctuations) AddSPE(timeBin, wave, fPMTGainFluctuationsPtr->GainFluctuation(1,fEngine));
-          else AddSPE(timeBin, wave);
-        }
+        if(timeBin < wave.size()) {AddSPE(timeBin, wave);}
       }
     }
 
@@ -182,10 +179,7 @@ namespace opdet {
         tphoton = ttsTime + auxphotons[j].Time - t_min + ttpb + fParams.CableTime;
         if(tphoton < 0.) continue; // discard if it didn't made it to the acquisition
         timeBin = std::floor(tphoton*fSampling);
-        if(timeBin < wave.size()){
-          if(fParams.MakeGainFluctuations) AddSPE(timeBin, wave, fPMTGainFluctuationsPtr->GainFluctuation(1,fEngine));
-          else AddSPE(timeBin, wave);
-        }
+        if(timeBin < wave.size()) {AddSPE(timeBin, wave);}
       }
     }
     // reflected light
@@ -198,10 +192,7 @@ namespace opdet {
         tphoton = ttsTime + auxphotons[j].Time - t_min + ttpb + fParams.CableTime;
         if(tphoton < 0.) continue; // discard if it didn't made it to the acquisition
         timeBin = std::floor(tphoton*fSampling);
-        if(timeBin < wave.size()){
-          if(fParams.MakeGainFluctuations) AddSPE(timeBin, wave, fPMTGainFluctuationsPtr->GainFluctuation(1,fEngine));
-          else AddSPE(timeBin, wave);
-        }
+        if(timeBin < wave.size()) {AddSPE(timeBin, wave);}
       }
     }
 
@@ -238,10 +229,7 @@ namespace opdet {
         tphoton = ttsTime + reflectedPhotons.first - t_min + ttpb + fParams.CableTime;
         if(tphoton < 0.) continue; // discard if it didn't made it to the acquisition
         timeBin = std::floor(tphoton*fSampling);
-        if(timeBin < wave.size()){
-          if(fParams.MakeGainFluctuations) AddSPE(timeBin, wave, fPMTGainFluctuationsPtr->GainFluctuation(1,fEngine));
-          else AddSPE(timeBin, wave);
-        }
+        if(timeBin < wave.size()) {AddSPE(timeBin, wave);}
       }
     }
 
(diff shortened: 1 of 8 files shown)


sbndcode/OpDetSim/PMTAlg/PMTGainFluctuations1Dynode_tool.cc
```diff
@@ -0,0 +1,72 @@
...
+
+  ~PMTGainFluctuations1Dynode() {}
+
+  //Returns fluctuated factor for SPR
+  double GainFluctuation(unsigned int npe, CLHEP::HepRandomEngine* eng) override;
+
+private:
+  //Configuration parameters
+  double fDynodeK;
+  unsigned int fGain;
```
ikatza, line 39: I believe this variable should be a double. Certainly that's the type of the input on the fcl file.

sbndcode/OpDetSim/PMTAlg/pmtgainfluctuations_config.fcl
```diff
@@ -0,0 +1,11 @@
+BEGIN_PROLOG
+
+FirstDynodeGainFluctuations:
+{
+  tool_type: "PMTGainFluctuations1Dynode"
+  DynodeK:                   0.75
+  VoltageDistribution:    [ 17.4, 3.4, 5.0, 3.33, 1.67, 1.0, 1.2, 1.5, 2.2, 3.0 ]
```
ikatza, line 7: Where do these values come from? If they're provisional, but say have to be calibrated, it would be good to add a TODO somewhere.

ikatza (changes requested):
All looks good and sensible. Please only comment on the `int/double` variable.

sbndcode/OpDetSim/DigiPMTSBNDAlg.cc
```diff
@@ -141,7 +145,10 @@ namespace opdet {
         tphoton = ttsTime + simphotons[i].Time - t_min + ttpb + fParams.CableTime;
         if(tphoton < 0.) continue; // discard if it didn't made it to the acquisition
         timeBin = std::floor(tphoton*fSampling);
-        if(timeBin < wave.size()) {AddSPE(timeBin, wave);}
+        if(timeBin < wave.size()){
+          if(fParams.MakeGainFluctuations) AddSPE(timeBin, wave, fPMTGainFluctuationsPtr->GainFluctuation(1,fEngine));
+          else AddSPE(timeBin, wave);
+        }
```
PetrilloAtWork, line 151: These two (or four) lines are very popular:
```suggestion
        AddSPEwithFluctuations(wave, timeBin);
```
They appear in the code four times. I **recommend** that they be put into a member function.

sbndcode/OpDetSim/DigiPMTSBNDAlg.cc
```diff
@@ -328,6 +349,17 @@ namespace opdet {
   }
 
 
+  void DigiPMTSBNDAlg::AddSPE(size_t time_bin, std::vector<double>& wave, double npe)
+  {
+    size_t max = time_bin + pulsesize < wave.size() ? time_bin + pulsesize : wave.size();
```
PetrilloAtWork, line 354: I **suggest** you use parentheses. The expression is correct, but I had to check a reference to verify the priorities between `?:`, `+` and `<`... I know I expressly forget operator priorities.
By the way, since you don't intend to change `max`, you can declare it `const`.

sbndcode/OpDetSim/PMTAlg/PMTGainFluctuations1Dynode_tool.cc
```diff
@@ -0,0 +1,72 @@
...
+  std::vector<double> fVoltageDistribution;
+
+  double fRefGain;
```
PetrilloAtWork, line 42: A comment that this is the gain on the first dynode is due.
Also, if possible, a more explicit name for this variable would also help: `fRefGain` was in my mind 10⁷, the nominal gain, which as such is also a reference gain.

PetrilloAtWork (commented):
The code looks correct to me.
I have left two recommendations and one suggestion, plus a few comments.
As such, none of them is a show stopper for merging but I invite you to consider them carefully.

I was under the impression that SBND was using `sim::SimPhotonsLite`, but I only see photons added one by one. I am surprised... but I did not see the full code, only the highlights from GitHub, so I have missed a lot.

sbndcode/OpDetSim/PMTAlg/pmtgainfluctuations_config.fcl
fjnicolas: Agree. I added some references and a TODO in the configuration fhicl file

sbndcode/OpDetSim/PMTAlg/PMTGainFluctuations1Dynode_tool.cc
fjnicolas: Changed with commit 
971f9b5

fjnicolas:
Changes after 971f9b5 (from suggestions from @PetrilloAtWork):
- Rename `fRegGain` variable to `fDynodeGain` in `opdet::PMTGainFluctuations1Dynode_tool.cc`
- Make `GainFluctuationsPset` a `fhicl::OptionalDelegatedParameter` and rename it to `GainFluctuationsParams` (in `DigiPMTSBNDSBNDAlg`)
- Use `art::ToolConfigTable<Config>`  for variable initialization in `opdet::PMTGainFluctuations1Dynode_tool.cc`
- Modified `AddSPE` function

lpaulucc (approved):
The code looks ok. I have no comments other than the ones already presented by Gianluca and Iker

sbndcode/OpDetSim/DigiPMTSBNDAlg.hh
```diff
@@ -132,7 +133,6 @@ namespace opdet {
     std::unique_ptr<opdet::PMTGainFluctuations> fPMTGainFluctuationsPtr;
 
     void AddSPE(size_t time_bin, std::vector<double>& wave); // add single pulse to auxiliary waveform
```
PetrilloAtWork, line 135:
```suggestion
    void AddSPE(size_t time_bin, std::vector<double>& wave) const; // add single pulse to auxiliary waveform
```

sbndcode/OpDetSim/PMTAlg/PMTGainFluctuations1Dynode_tool.cc
```diff
@@ -26,37 +29,56 @@ namespace opdet {
...
   double fDynodeK;
-  unsigned int fGain;
+  double fGain;
   std::vector<double> fVoltageDistribution;
 
-  double fRefGain;
+  double fDynodeGain;
 
-  unsigned int DynodeGain(unsigned int dynstage);
+  double DynodeGain(unsigned int dynstage);
```
PetrilloAtWork, line 68:
```suggestion
  double DynodeGain(unsigned int dynstage) const;
```

PetrilloAtWork (approved):
Thank you for your changes, Francisco.
I apologise I had not notice a coding detail in the first review; I am going to point it now, but since it's past the deadline, you can decide to just ignore it.

It is good to declare constant the methods which do not change the state of an object.
For example, `opdet::DigiPMTSBNDAlg::AddSPE()` does not; it changes a waveform, but that is passed as argument.
Another example is currently the fluctuator tool: effectively, `double PMTGainFluctuations1Dynode::GainFluctuation()` also does not change the state of the object and could be declared `const`. Here the choice is a bit more complicate, because to do that you would also need to change the base class interface, forcing the future implementations to be "const" too. The call is yours.

Anyway, formal approval here.

ikatza (approved):

pgreen135:
trigger build

PR #182, opened 2021-09-28
Fix for opDetDigitizerSBND

marcodeltutto:
Fixing a bug that was introduced with the update to art v3_09.

The `SimPhotons` Handle was saved to a new variable rather than the exiting class member. This caused no optical waveform to be created.


pgreen135:
trigger build SBNSoftware/sbncode#173

pgreen135:
Looks like this is working now in the CIs, thanks for the fix @marcodeltutto
DetSim...... | opdaq......... | ................................................... | std::vector<raw::OpDetWaveform>...................................... | ..5451

pgreen135 (approved):
Approve, although there appear to be some tab / whitespace issues?

PR #184, opened 2021-09-30
Update ci_tests.cfg

pgreen135:
required updates to ci_tests.cfg cpu/memory ranges following art v3_09


pgreen135:
trigger build

PR #186, opened 2021-10-11
Update MuonTrackProducer_module.cc

lynnt20:
bug: producer was ignoring tracks in tpc1 if a track existed in tpc0. changed an `else if` statement to an `if`, 
apologies for all the little bugs!


pgreen135:
trigger build

PR #204, opened 2021-11-03
Set SimChannel module as simdrift in main backtracker service settings

wforeman:
Setting this in the main inherited "sbnd_backtrackerservice" parameters, so individual modules don't have to worry about setting it.


marcodeltutto (approved):
Thanks @wforeman! Sorry I missed this when migrating to the new larg4

pgreen135:
trigger build

pgreen135 (approved):
CI's look good. Approve

PR #207, opened 2021-11-08
Getting CRT to work with the new LArG4

### LINKED ISSUE #137: CRT Geometry in v02_00
With the new geometry `v02_00`, the number of CRT auxiliary detectors returned by the GeometryService is 0. This likely has to do with the fact that the `physvol` name doesn't start with `volAuxDet`.

To quickly verify this behavior, with `sbndcode` set up:
```
python
import SBNDservices as serv
geom = serv.ServiceManager('Geometry')
geom.NAuxDets()
```
This returns:
- `142` with `sbndcode v09_24_00` (geometry `v01_05`).
- `0` with `sbndcode v09_28_00` (geometry `v02_00`).

### LINKED ISSUE #143: CRT Compatibility with the Refactored LArG4
Pull request #134 marks the switch to the refactored LArG4. The CRT is excluded by the detsim stage because it needs an extra module to parse the new larg4 output.

ilepetic:
This PR contains changes to sbndcode to allow the CRT to work with the new LArG4. This includes changes to the CRT detsim module and geometry. The geometry is changed to create individual logical volumes for each CRT scintillating strip (AuxDetSensitives). The strips are combined into an array (AuxDet).

Before this code is merged into develop, two other PRs must be completed:
1. https://github.com/LArSoft/larcorealg/pull/23
2. https://github.com/LArSoft/larsim/pull/78

This PR closes #137 and closes #143.


marcodeltutto:
Thank you @ilepetic! I just tried running it too and works all fine! One thing I have notices is that we now have many more `sim::AuxDetHit` data products saved. I see about 2188 `sim::AuxDetHit` separate data products after the `g4` stage, this is one product for each CRT sensitive volume. Am I doing something wrong or is this expected? Thanks!

Also, now that we have the CRT running, we should add this in the cosmics in-time and dirt workflows, and make sure we can make CRT-filtered samples. I'll open a separate issue to address these.

ilepetic:
The separate data products for each AuxDetSensitive are expected.

ilepetic:
sim::AuxDetHits are not necessary after the G4 stage, so they are now dropped.

marcodeltutto (changes requested):
Hi @ilepetic, thanks again! All looks good to me. The only thing I noticed is that you have generated `sbnd_v02_00.gdml` and `sbnd_v02_00_nowires.gdml` but not the equivalent ones with the overburden. Could you please also run
```bash
preparseGDML sbnd_v02_00_base.gdml -w -S DefaultWithShielding -o sbnd_v02_00_withshielding.gdml
``` 
and commit the two files with the overburden too? Thanks!

I am a bit concerned with the fact that we now have >1000 separate data-products saved in the event, one for each CRT strip. M concern is not about the file size as we are now dropping them, but on the fact that this makes the files quite hard to navigate now. Parsing the output of `eventdump.fcl` is now very hard; even after the products are dropped, they are still listed with a `?` under product size. Maybe there is a way to completely get rid of them I am not aware of? @jzennamo @wesketchum.

@ilepetic my understanding is that this is unavoidable because the new larg4 makes separate instances for each `volAuxDetSensitive`, and we cannot assign the same name to all the CRTs `volAuxDetSensitive` of the same type as it was before, because LArSoft geometry expects different names. Is my understanding correct? If so, maybe in the future we can talk to the LArSoft team and see what we can do about this, maybe LArG4 could merge all the `AuxDetHits` collections before placing them into the `Event`, just a thought.

Thank you again!

ilepetic:
GDML files with shielding have been added.

@marcodeltutto, that is my understanding as well.

marcodeltutto (approved):

pgreen135:
trigger build

pgreen135:
@ilepetic see build failures above

pgreen135:
trigger build LArSoft/larcorealg#23 LArSoft/larsim#78

pgreen135:
@ilepetic now builds correctly (missed the dependencies before!), but there are some failures at the detsim stage:

`%MSG-s ArtException:  PostEndJob 11-Nov-2021 06:51:05 CST ModuleEndJob
62: ---- EventProcessorFailure BEGIN
63:   EventProcessor: an exception occurred during current event processing
64:   ---- ScheduleExecutionFailure BEGIN
65:     Path: ProcessingStopped.
66:     ---- ProductNotFound BEGIN
67:       Found zero products matching all selection criteria
68:         C++ type: std::vector<sim::AuxDetSimChannel>
69:         Module label: 'genericcrt'
70:         Product instance name: ''
71:         Process name: (empty)The above exception was thrown while processing module sbndcode/CRT/CRTDetSim/crt run: 1 subRun: 0 event: 1
72:     ---- ProductNotFound END
73:     Exception going through path simulate
74:   ---- ScheduleExecutionFailure END
75: ---- EventProcessorFailure END
76: %MSG`

henrylay97:
Due to the nature of this PR we've triggered CI builds to use the sequential testing method (i.e. running all stages back to back rather than using old references as the inputs for each stage. These tests are not useful for showing product size differences but should prove end to end workflow (as the detsim error shows above). I'll post their results here when ready :)

henrylay97:
Sorry for the delay folks, weren't quite finished when I clocked off yesterday. All looks good from a sequential point of view (all stages run smoothly back-to-back (to and including reco2). Can find the builds here for [e20:prof](https://dbweb8.fnal.gov:8443/LarCI/app/ns:sbnd/build_detail/phase_details?build_id=sbnd_ci/7813&platform=Linux%203.10.0-1160.31.1.el7.x86_64&phase=ci_tests&buildtype=slf7%20e20:prof) and for [c7:prof](https://dbweb8.fnal.gov:8443/LarCI/app/ns:sbnd/build_detail/phase_details?build_id=sbnd_ci/7812&platform=Linux%203.10.0-1160.31.1.el7.x86_64&phase=ci_tests&buildtype=slf7%20c7:prof). 

The only issue these threw up were in memory or cpu usage, and that seems to be a calibration issue for the sequential tests. I ran an equivalent test against develop (which you can find [here](https://dbweb9.fnal.gov:8443/TestCI/app/ns:SBND/build_detail/phase_details?build_id=sbnd_ci/7822&platform=Linux%203.10.0-1160.31.1.el7.x86_64&phase=ci_tests&buildtype=slf7%20e20:prof) and it threw up the same results.

So from a CI point of view things things look pretty good. As long as we can confirm that the g4 product changes and the fcl file changes in the original non-sequential [e20 build](https://dbweb8.fnal.gov:8443/LarCI/app/ns:SBND/build_detail/phase_details?build_id=sbnd_ci/7807&platform=Linux%203.10.0-1160.31.1.el7.x86_64&phase=ci_tests&buildtype=slf7%20e20:prof) are to be expected (and they look as expected to me) then we're clear. 

I am in agreement with Marco that if there is anyway of cleaning up the product list from these ~1000 data products that would be extremely helpful when debugging!

pgreen135:
thanks @henrylay97, the g4 and fhicl changes make sense. I echo the comments that we should find a way to condense these `AuxDetHits` data products, the output from the g4 CI is difficult to read as there are so many of these.

pgreen135 (approved):

marcodeltutto:
Thanks @henrylay97 @pgreen135! After chatting with Ivan and Joseph, a separate issue will be opened to address the number of data products.

jzennamo:
trigger build

marcodeltutto:
I updated the GDML geometry to use only one physical volume for all the strips of the same type (so one physical volume for Y strips, one for X, and so on). This fixes #217.

pgreen135:
trigger build

Issue #130, opened 2021-08-02
GENIE Random Seed for CI Tests


henrylay97:
We've had a long running issue in the CI system where the `nucosmics_gen` test occasionally (ball park 1 in 20 runs) produces 2 neutrino interactions (2 `MCTruth` objects) rather than one. An example of the CI output can be found [here](https://dbweb8.fnal.gov:8443/LarCI/app/ns:sbnd/storage/docs/2021/07/27/stdout%234A2CME7.log) at the end of this log.

The [fcl](https://github.com/SBNSoftware/sbndcode/blob/develop/test/ci/sbnd_ci_nucosmics_gen_quick_test_sbndcode.fcl) used to run this test specifies "perEvent" and a specific seed so should theoretically always produce the same output. A few people have had brief looks at locating the issue but nothing detailed.

This issue has become more pressing now as two attempts to update reference events have failed due to the _new_ reference files being produced with 2 `MCTruth`'s and therefore failing the immediate [ci_test checks](https://dbweb8.fnal.gov:8443/LarCI/app/ns:sbnd/storage/docs/2021/07/31/stdout%23eCKxZSU.log) and therefore not being replaced.

Am assigning anyone I've already discussed this with at any point for book keeping.

vitodb:
By comparing the debug.log for the `nucosmics_gen` job used to generate the reference files and the corresponding job used to test the new reference I see there are differences in the seeds used by the random engine, I have the impression those seeds are set to `0`.

- Log for the reference generation
```
[...]
%MSG-d RANDOM:  ProcessEvent 31-Jul-2021 10:02:42 CDT run: 1 subRun: 0 event: 1
RNGservice::takeSnapshot_() of the following engine labels:
 | corsika:0:gen | corsika:0:pois |
[...]
%MSG-i NuRandomService:  CORSIKAGen:corsika@BeginModule  31-Jul-2021 10:02:42 CDT run: 1 subRun: 0 event: 1
Random seed for this event, engine 'corsika.gen': 760963931
%MSG
%MSG-i NuRandomService:  CORSIKAGen:corsika@BeginModule  31-Jul-2021 10:02:42 CDT run: 1 subRun: 0 event: 1
Random seed for this event, engine 'corsika.pois': 40438174
%MSG
%MSG-i CORSIKAGEN:  CORSIKAGen:corsika@BeginModule 31-Jul-2021 10:02:42 CDT  run: 1 subRun: 0 event: 1
[...]

PR #209, opened 2021-11-08
Turn off genie random seed flag

henrylay97:
This PR solves SBNSoftware/sbndcode#130 which has been a persistent issue with the CI. 

I did some further debugging on this today with the idea of opening a larsoft level issue. I established that alongside the original noted issue:
- the occasional occurrence of pileup events (visible in the CI output as a difference in the number of `MCTruth`)

There is also a more deep level random seed issue:
- the generated single event (even when the pileup doesn't occur) is non-identical (different energies & positions)

I found a section of [GENIEHelper](https://cdcvs.fnal.gov/redmine/projects/nugen/repository/revisions/master/entry/nugen/EventGeneratorBase/GENIE/GENIEHelper.cxx#L1592) that uses a fcl flag `UseHelperRndGen4GENIE` which is set to true by default. This parameter is meant to avoid issues with the random seed by storing it to re-initialise with a few lines later, however by deactivating it both seeding issues are solved and the CI produces the same event every time (I ran 20 iterations). 

I won't pretend to understand what exactly was going on with these different stored random number helpers in the relevant section but I _am_ confident this solves the persistent issue we've faced with this CI test.

The other changes in this commit are just extra changes I noticed whilst working on this (we don't need this lines they are resetting these parameters to the values they already hold from the inherited tables).

I've cast a wide net for reviews from anyone who's been involved in these discussion previously for completeness' sake.

### DIFF
--- a/test/ci/sbnd_ci_nucosmics_gen_quick_test_sbndcode.fcl
+++ b/test/ci/sbnd_ci_nucosmics_gen_quick_test_sbndcode.fcl
@@ -21,10 +21,7 @@ physics.producers.generator: {
   
   # ... and be as strict as possible with random seeds: GENIE is wild!
   Seed: 63213
-  
-  FluxCopyMethod: "DIRECT"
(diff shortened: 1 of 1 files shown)


pgreen135 (approved):
Thanks for digging into this Henry, approved from my side.

marcodeltutto (approved):
Thank you very much! Great to see this fixed

Issue #198, opened 2021-10-26
OpHitFinder does not work with (Daphne)XArapucas


rodralva:
This issue is related with the one pointed out by @ikatza  (#140) for FlashMatch. 
 
As far as i know, flash finder algorithms use [OpHits](https://nusoft.fnal.gov/larsoft/doxsvn/html/classrecob_1_1OpHit.html) as inputs. To compute the Ophits, algorithms look for peaks in the [raw simulated waveforms](https://nusoft.fnal.gov/larsoft/doxsvn/html/classraw_1_1OpDetWaveform.html).

In SBND we have 2 readout frequencies in the Optical System (500MHz/2ns for PMTs, 80MHz/12.5ns for Daphne XArapucas). The current SBND Optical Hit Finder [only considers](https://github.com/SBNSoftware/sbndcode/blob/develop/sbndcode/OpDetReco/OpHit/SBNDOpHitFinder_module.cc#L216) the default OpticalClock (500MHz) set for PMTs, and then gives this info to [larana algorithms](https://github.com/SBNSoftware/larana/blob/develop/larana/OpticalDetector/OpHitFinder/OpHitAlg.cxx#L29) calling the [RunHitFinder](https://github.com/SBNSoftware/sbndcode/blob/develop/sbndcode/OpDetReco/OpHit/SBNDOpHitFinder_module.cc#L230) function.

Simulated waveforms with daphne readouts have different number of bins, tick period... which results in miscalculations in the OpHits times and areas (PE). 

I'm working on this, modifying the SBND HitFinder so larana remains unchanged.

pgreen135:
Following https://github.com/SBNSoftware/sbndcode/pull/216 this has been (mostly) resolved. The remaining issues are documented in https://github.com/SBNSoftware/sbndcode/issues/225, so I will close this.

Issue #213, opened 2021-11-09
Unexplained OpHit / OpFlash product differences in CI tests following larana#14


pgreen135:
Several unexplained Reco1 product differences related to the OpHits and OpFlashes were picked up by the SBND CI tests following https://github.com/LArSoft/larana/pull/14 that made changes to the OpHitFinder for ICARUS (from Kazu). 

Full details from CI can be found here: https://dbweb8.fnal.gov:8443/LarCI/app/ns:sbnd/build_detail/phase_details?build_id=sbnd_ci/7640&platform=Linux%203.10.0-1160.31.1.el7.x86_64&phase=ci_tests&buildtype=slf7%20e20:prof 

@henrylay97 and I are investigating, have reached out to Kazu directly via slack and are awaiting a response. This needs to be understood/resolved before the production freeze.

henrylay97:
In order to ease with code checks of the many PRs currently being tested for MCP2021C we have updated the reference files to reflect this and a few other changes (the other changes are expected and understood. This issue acts as a reminder to ensure we see that the reverse changes happen when this is fixed. Or that the changes are understood.

Changes in nucosmics_reco1:
```676: Compare products sizes for nucosmics_reco1_test_sbndcode_Current.root output stream.
677: 
678: 
679: Check for differences in the size of data products
680: difference(s)
681: 
682: 47c47
683: < Reco1 | opflashtpc0 |  | art::Assns<recob::OpHit,recob::OpFlash,void> | 21749
684: ---
685: > Reco1 | opflashtpc0 |  | art::Assns<recob::OpHit,recob::OpFlash,void> | 21742
686: 49c49
687: < Reco1 | opflashtpc1 |  | art::Assns<recob::OpHit,recob::OpFlash,void> | 21418
688: ---
689: > Reco1 | opflashtpc1 |  | art::Assns<recob::OpHit,recob::OpFlash,void> | 21408
690: 51c51
691: < Reco1 | ophitarapuca |  | std::vector<recob::OpHit> | 533
692: ---
693: > Reco1 | ophitarapuca |  | std::vector<recob::OpHit> | 246
694: 54c54
695: < Reco1 | ophitpmt |  | std::vector<recob::OpHit> | 44033
696: ---
697: > Reco1 | ophitpmt |  | std::vector<recob::OpHit> | 44015
698: 101c101
699: < Reco1 | opflashtpc0 |  | art::Assns<recob::OpHit,recob::OpFlash,void> | 16614
[...]

PR #216, opened 2021-11-12
Deconvolution + HitFinder for XArapucas

rodralva:
This PR is related to the issue  #198 .

I added an electronics variable to the HitFinder and Deconvolution modules so we can distinguish daphne XArapucas from PMTs.
In the OpHitfinder side, I made an ad-hoc clock which is just a copy of the OpticalClock substituting the default optical frequency (500MHz) for the daphne readouts frequency (80MHz). The type of electronics then [decides which clock is passed](https://github.com/SBNSoftware/sbndcode/blob/bugfix/rodrigoa_hit_finder_xarapucas/sbndcode/OpDetReco/OpHit/SBNDOpHitFinder_module.cc#L232) to larana algorithm.
In this approach, deconvolved PMT and XArapuca waveforms are calculated separately.

### DIFF
--- a/sbndcode/OpDetReco/OpDeconvolution/Alg/OpDeconvolutionAlgWiener_tool.cc
+++ b/sbndcode/OpDetReco/OpDeconvolution/Alg/OpDeconvolutionAlgWiener_tool.cc
@@ -66,8 +66,8 @@ private:
 
   double fNormUnAvSmooth;
   double fSamplingFreq;
+  double fDaphne_Freq;
   size_t MaxBinsFFT;
-
   unsigned int NDecoWf;
 
   TF1 *fFilterTF1;
@@ -114,6 +114,7 @@ opdet::OpDeconvolutionAlgWiener::OpDeconvolutionAlgWiener(fhicl::ParameterSet co
   fOpDetDataFile = p.get< std::string >("OpDetDataFile");
   fFilter = p.get< std::string >("Filter");
   fElectronics = p.get< std::string >("Electronics");
+  fDaphne_Freq  = p.get< double >("DaphneFreq");
   fScaleHypoSignal = p.get< bool >("ScaleHypoSignal");
   fUseParamFilter = p.get< bool >("UseParamFilter");
   fFilterParams = p.get< std::vector<double> >("FilterParams");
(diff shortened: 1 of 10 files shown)


sbndcode/OpDetReco/OpDeconvolution/Alg/OpDeconvolutionAlgWiener_tool.cc
```diff
@@ -122,6 +124,7 @@ opdet::OpDeconvolutionAlgWiener::OpDeconvolutionAlgWiener(fhicl::ParameterSet co
 
   auto const clockData = art::ServiceHandle<detinfo::DetectorClocksService const>()->DataForJob();
   fSamplingFreq=clockData.OpticalClock().Frequency()/1000.;//in GHz
+  if (fElectronics=="Daphne") fSamplingFreq=80/1000.;//in GHz
```
pgreen135, line 127: The Daphne frequency should be a fcl parameter, rather than hardcoded

rodralva: You are absolutely right, it's a fcl parameter now.

pgreen135:
trigger build

pgreen135:
@rodralva in the CI I am seeing a sharp reduction in the number of OpHits for the arapucas, is this expected from this change?

```
615: < Reco1 | ophitarapuca |  | std::vector<recob::OpHit> | 50
616: ---
617: > Reco1 | ophitarapuca |  | std::vector<recob::OpHit> | 22
618: 93c93
619: < Reco1 | ophitarapuca |  | std::vector<recob::OpHit> | 31
620: ---
621: > Reco1 | ophitarapuca |  | std::vector<recob::OpHit> | 16
```

marcodeltutto (approved):
Thanks @rodralva, looks good to me!

jzennamo:
@pgreen135 is this not due to issue https://github.com/SBNSoftware/sbndcode/issues/213 ?

pgreen135:
@jzennamo no, the reference files have been updated since then -- so this change is new. If its unexpected though, this could be related (i.e. if something is broken with the ophitfinder)

fjnicolas (approved):

pgreen135:
trigger build

pgreen135:
I've re-run the CI's now that Kazu's PR has been reverted (https://github.com/SBNSoftware/sbndcode/issues/213), and the changes in the arapuca ophits remain. @rodralva what is the status of your tests relating to this?

rodralva:
The last deficit seen by CI test is expected (the 50% got me scared): there are XArapucas with slow readouts (12.5 ns tick), which are 176 channels, then we have 16 with apsaia readouts (2ns tick). For this first PR, I removed the Apsaia channels from OpHitFinder, because we don't have the true response (the data I took the single waveform from used daphne readouts, I only extrapolated the signal from 12.5ns tick to 2ns tick), and also because they are fewer so in this PR ophitarapuca only has Daphne channels. Apsaia channels are in the middle of the planes and account for ~10% of the total channels. As we only pass a vector of waveforms and one clock to larana, the only thing I can think of without changing larana is to make a 3rd ophit product (something like pmtophit, arapucaophit_daphne and arapucaophit_apsaia).

pgreen135:
thanks @rodralva for looking into this, I think this is all understood now and I will merge this (tomorrow). @rodralva and I have been discussing his comment above about the apsaia arapucas -- and issue will be opened related to this, and he will look at making these changes in larana (aiming for production next year)

PR #218, opened 2021-11-15
Reorganise dirt workflows to reduce fcl combinatorics

etyley:
With LArSoft v09_36_00  we can now run multiple G4 instances in a single fcl. This means that we can now reorganise the dirt overlay fcls so that a single G4 fcl can be used for all dirt samples (prodgenie or prodoverlay), reducing future maintenance burden etc.


pgreen135:
trigger build

marcodeltutto (approved):
Thank you @etyley!

PR #222, opened 2021-11-16
Feature/rodrigoa split s ers

rodralva:
Small update to the parameter that switches between Ideal and test(data driven) simulation of optical waveforms (detsim).
With this PR we can tune XArapucas and PMTs separately as each one has its own variable (replaced SinglePEmodel with ArapucaSinglePEmodel and PMTSinglePEmodel).

As the ideal response is outdated for XArapucas, I've switched the test response to default for the upcoming production. This will likely trigger warnings in the CI tests at ophit level from XArapucas. This is expected as the [test ser shape is thinner and presents overshooting](https://user-images.githubusercontent.com/70945401/128691609-24112902-21fb-4510-8b9c-0c022e55908a.png) (I expect ophits to decrease).

Ideal response for PMTs remains as default.


pgreen135:
trigger build

pgreen135 (approved):
CI's see small change, I think this makes sense.

PR #228, opened 2021-11-19
Fix for CRT detsim and reco

marcodeltutto:
With PR #207 we are using a CRT geometry that uses the same logical volumes for CRT strips of the same type. This caused the CRT detsim and reco to always pick the first CRT strip declared in the GDML. 

This PR fixes this problem by changing the logic in which the CRT geometry nodes are found. Now, the unique `AuxDet` volume is identified first, and then the code descends from it to find the strip (`AuxDetSensitive`).


jzennamo (approved):
I cannot say I am an expert on the ins and outs of the code but this looks reasonable and Marco says he did checks that behave as expected! Thanks Marco!

jzennamo:
trigger build

ilepetic (approved):

Issue #230, opened 2021-11-24
Need interim truth-matching solution/downstream EM shower rollup


wesketchum:
Because we are still waiting on the LArSoft patch for #202 , we need an interim solution for this for SBND (ICARUS did not make the jump to the new LArG4 for this release, so it shouldn't affect them).

As discussed in the Release Management meeting, we'll move ahead without this for this week/production testing, but we need something in place before launching most of our production samples.

I think our options are:
(1) Contribute directly on the LArSoft issue and push to have that resolved next week.
(2) Set `keepEMShowerDaughters` to True for SBND only, and perform a roll-up/truth-matching in the CAF creation.
(3) Some other clever solution?

wesketchum:
Pinging @henrylay97 @jzennamo @marcodeltutto @cjbackhouse @miquelnebot @pgreen135 etc. here so you are aware.

I hate to do it, but I think actually it makes sense to change `keepEMShowerDaughters` to True in SBND for as part of production testing for next week (still need downstream for CAFs). If that is sensible, @marcodeltutto or @henrylay97, can I ask you to get a PR in with that change made prior to a patch 2021C release going out this week?

jzennamo:
Could we create a second set of SBND fcl with `keepEMShowerDaughters == true` that we can use in test samples before we launch production?

henrylay97:
So there already exists a fcl `g4_no_shower_rollup.fcl` which lives [here](https://github.com/SBNSoftware/sbndcode/blob/develop/sbndcode/JobConfigurations/standard/g4/g4_no_shower_rollup.fcl) which runs standard g4 without rollup.

On [this](https://github.com/SBNSoftware/sbndcode/tree/feature/hlay_no_rollup) branch I've added an equivalent for sce enabled g4. 

I'm currently running some tests on the standard CI to do a basic assessment of memory and time differences.

henrylay97:
First set of CI results. Running the nu+cosmics regression test. The g4 stage (processing an artroot file with 2 BNB+Cosmic events).

[**v09_37_01**](https://dbweb8.fnal.gov:8443/LarCI/app/ns:sbnd/build_detail/test_details?build_id=sbnd_ci/8081&platform=Linux%203.10.0-1160.45.1.el7.x86_64&buildtype=slf7%20e20:prof&phase=ci_tests&test=nucosmics_g4_quick_test_sbndcode)
Elapsed time: 660.29 (s)
[...]

PR #231, opened 2021-11-26
Rollup fcls for MCP2021C testing

henrylay97:
This PR adds the fcls required to test rollup being turned off, and further turning off mcreco as discussed in #230.


pgreen135:
are these additional fhicls sufficient for production testing? there are various other workflows that inherit the standard_g4 fhicls that wouldn't pick this up -- e.g. the dirt workflows, PDS ones, commissioning ones

henrylay97:
This is a really good point, sorry! Has anyone got a list of the fcls intended to be used in this production?

pgreen135:
discussing with @henrylay97, the plan is to merge this into today's release as is to enable testing of the core workflows on Monday. Then next week there will be a followup patch with the full set of configurations pending the results from the testing / input from US people (some workflows may need MCReco etc.).

jzennamo:
Thank you, this was exactly what we needed, I have started testing now and I can work on getting patch fcls in place based on these!

PR #233, opened 2021-12-01
CAF tests for the CI system

henrylay97:
_N.B. This PR depends on [this](https://cdcvs.fnal.gov/redmine/projects/lar_ci/repository?utf8=%E2%9C%93&rev=feature%2Fhlay_caf_tests) branch in `lar_ci` and so this will need approving and merging before the `trigger build` command can be used. @vitodb if you wouldn't mind taking a look at this that would be great._

This PR introduces the CAF stage to the standard regression tests. These are the simple tests run on **every** PR and push to develop. At the moment we test [gen, g4, detsim, reco1, reco2, anatree]. This PR _replaces_ the anatree stage with a caf stage. I don't think the anatree is widely used anymore and the CI test for it only checks that it can be built and not anything inside it. I have retained the functionality for the anatree test but replaced in the suites that are automatically run.

There are huge benefits to a caf test. The obvious is that we test the CAFMaker infrastructure which is currently not covered by the CI at all (other than it compiles). The real extra benefit compared to the other stages is that in the simple inexpensive test we are able to test physics quantities without running the full validation. Thanks to the lovely `diff_cafs` script we will see if any value in the CAF has changed.

Something I haven't addressed in this PR is that the production workflow is currently using `*_sce` fcls. We should probably swap to this going forward (in place of the `standard_*.fcl` we currently use) but this will be addressed separately.

The fcl use for the CI test isn't as simple as the other stages so for completeness I'll document why:
- There doesn't currently exist a file that runs the flux & genie weighting system without space charge effects, I have introduced this fcl in this PR.
- To use the weighting functionality without making the run time too long for a regression test I've then reduced the multisims to 1 universe, thus testing the mechanics without wasting time.
- We can't run weighting on the prodsingle version of the test so it has to inherit from a different fcl.

Examples of this test in action can be found [with](https://dbweb8.fnal.gov:8443/LarCI/app/ns:sbnd/build_detail/phase_details?build_id=sbnd_ci/8102&platform=Linux%203.10.0-1160.45.1.el7.x86_64&phase=ci_tests&buildtype=slf7%20e20:prof) and [without](https://dbweb8.fnal.gov:8443/LarCI/app/ns:sbnd/build_detail/phase_details?build_id=sbnd_ci/8103&platform=Linux%203.10.0-1160.45.1.el7.x86_64&phase=ci_tests&buildtype=slf7%20e20:prof) warnings. To run cleanly this test requires SBNSoftware/sbnanaobj#33

### DIFF
--- a/test/ci/sbnd_ci_nucosmics_caf_quick_test_sbndcode.fcl
+++ b/test/ci/sbnd_ci_nucosmics_caf_quick_test_sbndcode.fcl
@@ -1,0 +1,8 @@
+#include "cafmakerjob_sbnd.fcl"
+
+services.NuRandomService.policy: "perEvent"
+
+# Have to specify the fcl name to match what the CI expects,
+# currently no way of controlling this from the command line
+
+physics.producers.mycafmaker.CAFFilename: "nucosmics_caf_test_sbndcode_Current.caf.root"
--- a/test/ci/sbnd_ci_nucosmics_caf_quick_test_sbndcode.fcl
+++ b/test/ci/sbnd_ci_nucosmics_caf_quick_test_sbndcode.fcl
@@ -2,7 +2,7 @@
 
 services.NuRandomService.policy: "perEvent"
 
-# Have to specify the fcl name to match what the CI expects,
+# Have to specify the caf name to match what the CI expects,
 # currently no way of controlling this from the command line
 
 physics.producers.mycafmaker.CAFFilename: "nucosmics_caf_test_sbndcode_Current.caf.root"
--- a/test/ci/sbnd_ci_nucosmics_caf_quick_test_sbndcode.fcl
+++ b/test/ci/sbnd_ci_nucosmics_caf_quick_test_sbndcode.fcl
@@ -1,4 +1,4 @@
-#include "cafmakerjob_sbnd.fcl"
+#include "cafmakerjob_sbnd_genie_and_fluxwgt.fcl"
 
 services.NuRandomService.policy: "perEvent"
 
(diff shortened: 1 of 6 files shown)


test/ci/sbnd_ci_nucosmics_caf_quick_test_sbndcode.fcl
```diff
@@ -0,0 +1,12 @@
+#include "cafmakerjob_sbnd_genie_and_fluxwgt.fcl"
+
+services.NuRandomService.policy: "perEvent"
+
+# Have to specify the fcl name to match what the CI expects,
```
cjbackhouse, line 5:
```suggestion
# Have to specify the caf name to match what the CI expects,
```

cjbackhouse (approved):
Very happy this is going to exist. I don't see the string `diff_cafs` anywhere in here. Was that already implemented in a previous PR, and this is just hooking that ability up to some actual test fcls?

test/ci/sbnd_ci_nucosmics_caf_quick_test_sbndcode.fcl
henrylay97: fab thanks!

henrylay97:
Hi Chris, thanks for the prompt review. You're right, the actual functionality for running the CAF test has been added to the script in lar_ci that runs the regression tests. The relevant commit can be found [here](https://cdcvs.fnal.gov/redmine/projects/lar_ci/repository/revisions/898e038dcb3cb666b46731733889be0ced1a5485/diff/bin/ci_regression_test_template.sh) (note the added function `compare_cafs`) if you're interested.

vitodb (approved):
This looks good!
I just merged Henry's lar_ci feature branch, so this PR has what is needed from the CI infrastructure to be able to run

PR #237, opened 2021-12-08
Change caf module label to cafmaker not mycafmaker

henrylay97:
Production team are moving towards using a common set of scripts in sbnutil. One issue I came across during this migration is that ICARUS use the module name `cafmaker` whilst we use `mycafmaker` for CAFMaker. This PR moves us to use the same as ICARUS such that the scripts don't need extra logic for sbnd files. 

I don't think any other changes should be needed as cafs aren't ever used as art inputs via the module name but I'm willing to be corrected by those who know more.


cjbackhouse (approved):
Yes please!

pgreen135:
trigger build

PR #240, opened 2021-12-10
Patch for CRT error where AuxDets can't be found

henrylay97:
**Very** intermittent error found by @jzennamo. Also thanks to @marcodeltutto and @wesketchum for help with debugging.

```
---- ChannelMap BEGIN
  Can't find AuxDet for position (132.794,-376.746,-8.4112)
  The above exception was thrown while processing module GenericCRT/genericcrt run: 1 subRun: 0 event: 84
---- ChannelMap END
```

This error is being thrown in the [ChannelMapAlg](https://github.com/LArSoft/larcorealg/blob/develop/larcorealg/Geometry/ChannelMapAlg.cxx) but this PR overrides the relevant functions in sbndcode's ChannelMapSBNDAlg.

Instead of throwing an exception the same error message is now outputted as a debugging message and the function returns `UINT_MAX` as the channel ID.

A catch is added at the detsim stage and in a detsim analysis module to skip channels with the ID `UINT_MAX`.

The underlying bug is worth investigating further and @ilepetic can probably comment further.

Due the rare nature of this bug I haven't been able to consistently reproduce it to test these changes on, however I have made some tests of the detsim stage where I forcibly return `UINT_MAX` and the catch works fine.

### DIFF
--- a/sbndcode/CRT/CRTAna/CRTDetSimAna_module.cc
+++ b/sbndcode/CRT/CRTAna/CRTDetSimAna_module.cc
@@ -328,6 +328,10 @@ namespace sbnd {
     for (auto& adsc : *channels) {
       // Get the IDEs from the Aux Det channels
       const geo::AuxDetGeo& adGeo = fGeometryService->AuxDet(adsc.AuxDetID());
+
+      if(adsc.AuxDetSensitiveID() == UINT_MAX)
+        continue;
+
       const geo::AuxDetSensitiveGeo& adsGeo = adGeo.SensitiveVolume(adsc.AuxDetSensitiveID());
       std::string stripName = adsGeo.TotalVolume()->GetName();
       stripName.append("_0");
--- a/sbndcode/CRT/CRTAna/CRTDetSimAna_module.cc
+++ b/sbndcode/CRT/CRTAna/CRTDetSimAna_module.cc
@@ -329,8 +329,11 @@ namespace sbnd {
       // Get the IDEs from the Aux Det channels
       const geo::AuxDetGeo& adGeo = fGeometryService->AuxDet(adsc.AuxDetID());
 
-      if(adsc.AuxDetSensitiveID() == UINT_MAX)
+      if(adsc.AuxDetSensitiveID() == UINT_MAX) {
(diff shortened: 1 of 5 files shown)


sbndcode/CRT/CRTDetSim_module.cc
```diff
@@ -156,6 +156,10 @@ void CRTDetSim::produce(art::Event & e) {
   for (auto& adsc : *channels) {
     const geo::AuxDetGeo& adGeo =
         geoService->AuxDet(adsc.AuxDetID());
+
+    if(adsc.AuxDetSensitiveID() == UINT_MAX)
```
marcodeltutto, line 160: Can you please also add a LOG_WARNING here with something like "Skipping AuxDetSensitiveID with bad channel number." This happens rarely so it won't clog our log files, but at least we see it and remember it's something we'll have to fix.

sbndcode/CRT/CRTAna/CRTDetSimAna_module.cc
```diff
@@ -328,6 +328,10 @@ namespace sbnd {
     for (auto& adsc : *channels) {
       // Get the IDEs from the Aux Det channels
       const geo::AuxDetGeo& adGeo = fGeometryService->AuxDet(adsc.AuxDetID());
+
+      if(adsc.AuxDetSensitiveID() == UINT_MAX)
```
marcodeltutto, line 332: Similarly here

marcodeltutto (changes requested):
Looks great, thanks so much @hlay! I just have a couple of comments.

Could you also remove this exception [here](https://github.com/SBNSoftware/sbndcode/blob/c569938ccabe635bcae6382832d65898c96ffe2c/sbndcode/CRT/CRTChannelMapAlg.cxx#L125) and add a debug message as you did in `ChannelMapSBNDAlg`? Otherwise, if someone uses `AuxDetGeometryCore` that will throw an exception.

henrylay97:
Thanks @marcodeltutto, all great suggestions. I've followed up in the above commits :+1:

pgreen135:
trigger build

pgreen135:
trigger build

marcodeltutto (approved):
Looks good, thank you!

ilepetic (approved):

PR #246, opened 2022-01-11
More CRT Segfaults

henrylay97:
This further patches CRT segfaults related to those addressed in #240. These segfaults originated in the introduction of the refactored larg4 for the CRT (#207). These are "rare" segfaults, hence they were discovered in production where the volume of events was high enough to see this occur regularly.

This has been 'tested' in production by using a tarred version of the code. It fixes the issue we were seeing.

### DIFF
--- a/sbndcode/CRT/CRTDetSim_module.cc
+++ b/sbndcode/CRT/CRTDetSim_module.cc
@@ -154,6 +154,12 @@ void CRTDetSim::produce(art::Event & e) {
 
   // Loop through truth AD channels
   for (auto& adsc : *channels) {
+    if(adsc.AuxDetID() == UINT_MAX) {
+      mf::LogWarning("CRTDetSim") << "AuxDetSimChannel with ID: UINT_MAX\n"
+                                  << "skipping channel...";
+      continue;
+    }
+
     const geo::AuxDetGeo& adGeo =
         geoService->AuxDet(adsc.AuxDetID());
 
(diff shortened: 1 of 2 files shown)


ilepetic (approved):

marcodeltutto (approved):
Thanks, @henrylay97!

I confirm that this should fix the segfault. I was able to reproduce the issue and `gdb` traced it back to [this line](https://github.com/SBNSoftware/sbndcode/blob/803f34de463ee4a8d18385c0589d536cb60e6262/sbndcode/Geometry/ChannelMapSBNDAlg.cxx#L53):

```c
geo::AuxDetGeo const& adg = auxDets[auxDetIdx];
```
which is exactly what is being fixed by this PR.

--------------------------
`gdb` output:
```bash
[...]
Skip SimEnergyDeposit in: LArG4DetectorServicevolCryostat
Skip SimEnergyDeposit in: LArG4DetectorServicevolXArapuca

Thread 1 "lar" received signal SIGSEGV, Segmentation fault.
geo::ChannelMapSBNDAlg::NearestSensitiveAuxDet (this=<optimized out>, point=0x7ffffffe1540, auxDets=..., tolerance=0.0001)
    at /scratch/workspace/sbndcode-release-build/BUILDTYPE/prof/QUAL/e20/label1/swarm/label2/SLF7/temp/srcs/sbndcode/sbndcode/Geometry/ChannelMapSBNDAlg.cxx:53
```

henrylay97:
Ah wonderful, very grateful you were able to actually establish a definite test @marcodeltutto

PR #250, opened 2022-01-24
CRT Plane Index and Commissioning Samples

marcodeltutto:
This PR:
- From Michelle: adds a `GetPlaneIndex` function in `CRTCommonUtils` to retrieve the CRT plane index. This is also saved in the CRT Hit dataproduct.
- Fixes the CRT plane names: Front, Back, Left, Right -> South, North, West, East
- Makes sure that the commissioning detsim fhicl saves the full optical waveforms 

This PR is also needed to generate the commissioning samples for the current 2021C production.

### DIFF
--- a/sbndcode/CRT/CRTChannelMapAlg.cxx
+++ b/sbndcode/CRT/CRTChannelMapAlg.cxx
@@ -91,8 +91,6 @@ namespace geo {
       size_t& ad,
       size_t& sv) const {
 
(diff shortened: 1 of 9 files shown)


pgreen135:
trigger build

mstancar:
Looks good, simple change

pgreen135:
Disregard the CI failures, they are I think due to the incompatibility between this version and the current larsoft (art updates etc.). I will make a tag for this later today

mstancar (approved):

PR #251, opened 2022-01-25
Adding fcl to drop mcreco for intime cosmics

jzennamo:
Running MCReco for in-time cosmics isn't needed by physics groups and it is very very slow to run without shower-rollup.


marcodeltutto (approved):

jzennamo:
I was able to walk through the full production workflow in finite time which gives me confidence that this is working. I'll let @henrylay97 comment further if this solves all our issues.

henrylay97 (approved):
This looks good to me, thank you for doing this @jzennamo. I have verified it speeds things up offline in the way I would expect. @pgreen is there chance of a production patch with this in so we can run it in production. Thanks!

pgreen135:
@henrylay97 is this ready for a patch release today?

pgreen135:
@henrylay97 is this ready for a patch release today?

henrylay97:
I believe so if you have time :)

PR #259, opened 2022-02-18
Feature/lynnt_pmt_trig pmtTrigger producer updates

lynnt-uchicago:
improvements to memory allocation, removed reference of ophit and opdaq in the fcl file


eyandel (approved):
Looks good! Uncommenting the "double fBaseline = 8000.0; //baseline ADC (set in simulation)" line may lead to some "unused variable" errors as that is not currently being used and was left there in case we changed the pmt response and had to find the baseline later. You might want to comment it again before the automatic compilation checks.

lynnt-uchicago:
got it! re-commented the baseline line

pgreen135:
trigger build

PR #264, opened 2022-03-16
CRUMBS

henrylay97:
The functionality in sbndcode to run the new CRUMBS tool which lives in sbncode.

- CRUMBS is added as a producer to the standard and standard+sce caf workflows
- SCE versions of the crt hit and track matching are added as they are needed as inputs
- sbndcode is pointed to a new version of sbnd_data (v1_12_00) containing the weights file for CRUMBS's BDT

Note this PR assumed that #255 and #256 had already been merged into `release/SBN2021C`. Once this occurs then these extra commits should be removed from the diff.

Links to main PR SBNSoftware/sbncode#250

### DIFF
--- a/sbndcode/JobConfigurations/standard/caf/cafmakerjob_sbnd.fcl
+++ b/sbndcode/JobConfigurations/standard/caf/cafmakerjob_sbnd.fcl
@@ -17,6 +17,7 @@
 #include "showerselectionvarsproducer.fcl"
 #include "showercosmicdistanceproducer.fcl"
 #include "sbn_pid.fcl"
+#include "crumbs_producer.fcl"
 
 #include "crthitconverter_producer.fcl"
 #include "crtmatch_producer.fcl"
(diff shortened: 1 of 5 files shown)


etyley (approved):
I can't really comment on the CRT changes, but the fcl/xml/ups changes all look reasonable to me

henrylay97:
Perfect, thanks Ed! Yeah those CRT changes are hoovered up from #256 (which has been approved) and shouldn't be in this diff that's my bad, they will disappear as soon as that merged!

sbndcode/JobConfigurations/standard/caf/cafmakerjob_sbnd.fcl
```diff
@@ -17,6 +17,7 @@
 #include "showerselectionvarsproducer.fcl"
 #include "showercosmicdistanceproducer.fcl"
 #include "sbn_pid.fcl"
+#include "crumbs_producer.fcl"
```
absolution1, line 20: If this fcl lives in an SBN repo then may want to include 'sbn' in the name to indicate that it isn't in central larsoft

absolution1 (approved):
Looks good to me.
I left want comment about labelling but it can be addressed later.

PR #273, opened 2022-03-30
Add fcl testing framework based on icarus fcl testing framework.

hgreenlee:
Copy icaruscode fcl testing unit test into sbndcode.


pgreen135:
@hgreenlee why is this needed for 2022A? it will require a patch release to include it; or can this just be merged into develop?

pgreen135:
@henrylay97 is this doing the same / similar to the already existing fcl checks you added? or is this different?

henrylay97:
Unless it is desperately needed for 2022A in which case you may need someone else to take a look, not sure if it is tho?

pgreen135:
Closing, replaced with https://github.com/SBNSoftware/sbndcode/pull/275

PR #283, opened 2022-05-16
Run CRUMBS as standalone producer

henrylay97:
This adds a fcl for running CRUMBS as a standalone producer. As mentioned by issue SBNSoftware/sbncode#164 CRUMBS and other high-level reco / analysis tools are run as part of the CAF fcl stage. In future the idea is to move most of these producers into the reco2 stage but in the meanwhile it is useful to have a fcl that demonstrates how to run CRUMBS on artroot files if you don't want to use the CAFs.

### DIFF
--- a/sbndcode/JobConfigurations/standard/ana/crumbs_sbnd.fcl
+++ b/sbndcode/JobConfigurations/standard/ana/crumbs_sbnd.fcl
@@ -5,8 +5,6 @@
 #include "sbn_crumbs_producer.fcl"
 
 #include "rootoutput_sbnd.fcl"
-#include "geometry_sbnd.fcl"
-#include "messages_sbnd.fcl"
 #include "services_sbnd.fcl"
 
 process_name: CRUMBS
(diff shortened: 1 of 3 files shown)


sbndcode/JobConfigurations/standard/ana/crumbs_sbnd.fcl
```diff
@@ -0,0 +1,48 @@
...
+
+process_name: CRUMBS
+
+services:
+{
+   TFileService: { fileName: @local::sbnd_tfileoutput.fileName }
+   Geometry: @local::sbnd_geo
+   ExptGeoHelperInterface:    @local::sbnd_geometry_helper
+   @table::sbnd_services
+   @table::sbnd_resourcemonitorservices
```
absolution1, line 20: Do you need all of these?  Presumably `sbnd_services` should fold in most of them?

henrylay97:
Thanks @absolution1 turns out you were correct on both counts, who would've guessed?

fjnicolas:
trigger build

PR #286, opened 2022-06-07
Modify reco fcls so there's an option for not using the OpT0Finder

ikatza:
Just as the title says.

I'm open to suggestions about a neater way to achieve this.


pgreen135:
The no flash matching fhicls should also disable the simplified flash matching -- otherwise the name is rather confusing!

marcodeltutto (approved):

pgreen135 (approved):
Thanks for updating the names @ikatza

fjnicolas:
trigger build

fjnicolas:
Hi @ikatza, unit tests failed for the `standard_reco1reco2_no_opt0finder_sbnd_lite.fcl` fhicl. Could you update the header to the `no_opt0finder` version? Thanks!

ikatza:
Sorry for the oversight, I've pushed the rename.

fjnicolas:
trigger build

fjnicolas:
@ikatza can you have a look to the CIs? The `reco2` sequence is not well defined. I suggest defining it as `physics.reco2: [ @sequence::physics.reco2_no_opt0finder, opt0finder ]` outside the physics block.

fjnicolas:
trigger build

PR #288, opened 2022-06-15
CRT Hit Timings

henrylay97:
I presented these changes at today's SBND Reco meeting (see [docDB#26626](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=26626)). Currently we simulate two key timing effects in the CRT strips. 

- Propagation delay (light takes time to travel along the fibre).
- Time walk (bigger pulses record an earlier time due to reaching the threshold earlier).
but we don't account for any of this in reconstruction.

This PR does the following:

- Corrects the propagation speed value (currently simulating 10x the speed it should)
- Adds a correction method to the CRT Hit Reconstruction to account for these two effects using reconstruction information.
- The relevant fcl parameters are all inherited directly from the ones used in detsim.
- Finally, [this](https://github.com/SBNSoftware/sbndcode/commit/3647e3c4c6a82dfa85195bc428d940cb899ae703) commit changes a std::cout to a LogInfo dump, is that okay @marcodeltutto?


ilepetic (approved):

marcodeltutto (approved):
Thanks @henrylay97! Yes, sorry about the couts (leftovers!).

PR #294, opened 2022-08-12
Update ci_tests.cfg

henrylay97:
Reflecting current memory usage.

Have also updated the seq tests to have more sensible resource requirements. This is because they run more events than the quick tests (5 vs. 2) and we have never tuned the values. We've recently started using this test more so would be helpful to have these limits added!


fjnicolas (approved):
thanks @henrylay97 for tuning the values!

Issue #62, opened 2020-10-30
[MCP2020A] Issue With OpHits


gputnam:
While testing the production for MCP 2020A, we encountered an issue where the simple flash matching (Iker's) was not producing output. We traced this to an issue where all of the OpHit's from PMT's are produced with "0" filled in every variable in each hit as seen by Iker's module.

From my view, it seems like we should hold off on production until we can confirm the integrity of the OpHits.

PR #296, opened 2022-09-07
Feature/eyandel trigger

eyandel:
- Made some modules to combine CRT and PMT simulated fragments into one event. 
- Put CRT software trigger metrics module @mstancar made for sbndaq into sbndcode (needs to include [sbnobj PR #62](https://github.com/SBNSoftware/sbnobj/pull/62) as well). 
- Make "T1 reset" events for CRT boards with no hits in an event to avoid missing fragments.


lynnt-uchicago:
Hi Erin!! thanks for all the work of adding/removing files!! For the MetricsProducer module, are the pmt metrics more or less copy/pasted over? from a quick look it didn't look like you changed anything about the structure. is the eventual plan to remove the separate PMT software trigger module and only use the MetricsProducer module?

sbndcode/Trigger/run_metricproducer.fcl
```diff
@@ -0,0 +1,51 @@
...
+
+ stream1: [ out1 ]
+
+ trigger_paths: [ mode0 ]
+
+ end_paths: [ stream1]
+
+}
+
+physics.producers.SoftwareTriggerMetric.BeamWindowStart: 1650000
```
lynnt-uchicago, line 49: if the eventual plan is to have both the CRT and the PMTs in this same "SoftwareTriggerMetric" producer, maybe we need to specify CRTBeamWindow and PMTBeamWindow?

sbndcode/Commissioning/HitDumper_module.cc
```diff
@@ -418,6 +426,7 @@ void Hitdumper::reconfigure(fhicl::ParameterSet const& p)
   freadOpHits        = p.get<bool>("readOpHits",true);
   freadpmtTrigger    = p.get<bool>("readpmtTrigger",true);
   freadpmtSoftTrigger= p.get<bool>("readpmtSoftTrigger",true);
+  freadcrtSoftTrigger= p.get<bool>("readcrtSoftTrigger",false);
```
henrylay97, line 429: Sanity check... pmt soft trigger defaults true, crt defaults false... Deliberate?

sbndcode/Commissioning/fcls/detsim_comm_fhicldump.out
```diff
@@ -0,0 +1,1043 @@
+# Produced from 'fhicl-dump' using:
```
henrylay97, line 1: Naive question again, what's the reason for committing this dump output?

sbndcode/Commissioning/HitDumper_module.cc
```diff
@@ -252,6 +254,9 @@ class Hitdumper : public art::EDAnalyzer {
   int    _pmtSoftTrigger_nAboveThreshold;    /// number of individual PMTs above ADC threshold (fcl) during the beam spill
   // std::vector<sbnd::trigger::pmtInfo> _pmtSoftTrigger_pmtInfoVec; /// vector of PMT information 
 
+  // CRT software trigger variables
+  int    _crtSoftTrigger_hitsperplane[7];       ///< Number of (very low level) CRT hits per plane
```
henrylay97, line 258: Being picky here... Even though this will always be 7, I'm not a fan of hard coded numbers. Personally I would code this as a std::vector which is resized by using the geom service to get the number of planes.

sbndcode/Commissioning/HitDumper_module.cc
```diff
@@ -893,6 +902,24 @@ void Hitdumper::analyze(const art::Event& evt)
...
 
+  //
+  // CRT Software Trigger
+  //
+  if (freadcrtSoftTrigger){
+    art::Handle<sbndaq::CRTmetric> crtSoftTriggerHandle;
+    if (evt.getByLabel(fcrtSoftTriggerModuleLabel, crtSoftTriggerHandle)){
+      const sbndaq::CRTmetric &crtSoftTriggerMetrics = (*crtSoftTriggerHandle);
+      ResetCrtSoftTriggerVars();
+      for (int i=0; i<7; i++){
```
henrylay97, line 913: as with earlier comment this would allow you to avoid this hard coding too

sbndcode/Commissioning/HitDumper_module.cc
```diff
@@ -1456,6 +1487,12 @@ void Hitdumper::ResetPmtSoftTriggerVars(){
   _pmtSoftTrigger_nAboveThreshold = 0;
 }
 
+void Hitdumper::ResetCrtSoftTriggerVars(){
+  for (int i=0; i<7; i++){
```
henrylay97, line 1491: and here, crops up in other modules as well, I think you're probably going to ignore me (as I probably would to ;) ) so I won't comment each one.

sbndcode/Trigger/ArtdaqFragmentProducer_module.cc
```diff
@@ -0,0 +1,611 @@
...
+        std::string stripName = fCrtGeo.ChannelToStripName(channel);
+        std::string tagger = fCrtGeo.GetTaggerName(stripName);
+        taggers[feb_data->Mac5()] = tagger;
+
+        T0s[feb_data->Mac5()][feb_hits_in_fragments[feb_data->Mac5()]] = feb_data->Ts0();
+        T1s[feb_data->Mac5()][feb_hits_in_fragments[feb_data->Mac5()]] = feb_data->Ts1();
+        for (int i_adc = 0; i_adc<32; i_adc++){
+          uint16_t adc = 0;
+          if (feb_data->ADC()[i_adc] > 4089){
+            adc = 4089;
```
henrylay97, line 289: Out of interest why 4089? Would threshold not be 4095?

sbndcode/Trigger/MetricProducer_module.cc
```diff
@@ -0,0 +1,509 @@
...
+        // normal fragment
+        size_t beamFragmentIdx = -1;
+        for (auto frag : *handle){
+          beamFragmentIdx++;
+          if (frag.type()==sbndaq::detail::FragmentType::BERNCRTV2) {
+            num_crt_frags++;
+          	if (fcrt_metrics){analyze_crt_fragment(frag);}
+          }
+
+
```
henrylay97, line 206: Quite a lot of places, like here, where cleaning up the formatting would make it a lot more readable, again not really something for now.

henrylay97 (approved):
Looks fab Erin, thanks for doing this. A lot of this is outside of my scope. I understand what you're doing with the combining of the simulated fragment producers and the production of a CRT fragment regardless of activity, all looks great. 

I've included a few comments but they're all totally naive and ignorable / stylistic so from my side its good to go.

eyandel:
> Hi Erin!! thanks for all the work of adding/removing files!! For the MetricsProducer module, are the pmt metrics more or less copy/pasted over? from a quick look it didn't look like you changed anything about the structure. is the eventual plan to remove the separate PMT software trigger module and only use the MetricsProducer module?

Yes everything is pretty much just copied over. I think ideally we will move to just using the one module so that we can save time by not opening fragments multiple times. This is just the first working attempt at that for now. It will need a bit more work to get the structure to fit together more efficiency I think.

sbndcode/Trigger/run_metricproducer.fcl
eyandel: I agree. I believe they are currently separate but not named well due to the fact that I just took them from their separate codes. This will be one of the things we should update when we start improving this module.

sbndcode/Commissioning/HitDumper_module.cc
eyandel: An unfortunate consequence of multiple people writing the code and me stitching it together. I think I defaulted the new one to true since the products for it were not in sbnobj yet and I didn't want to break anything if someone updates and doesn't know these new products are there.

sbndcode/Commissioning/fcls/detsim_comm_fhicldump.out
eyandel: it's deleted now

sbndcode/Commissioning/HitDumper_module.cc
eyandel: I agree. However, I mostly was just moving stuff over from what Michelle made in sbndaq and she hardcoded this. I think we should fix it there first before changing it here, since I want this to be basically as close to the daq version as possible. But yes this will hopefully be changed in the future.

sbndcode/Commissioning/HitDumper_module.cc
eyandel: Hopefully these will all be changed eventually! so notes about where we need to change this are helpful thank you.

sbndcode/Trigger/ArtdaqFragmentProducer_module.cc
eyandel: data is at 4089 for some reason. I don't know why but that's what we see.

sbndcode/Trigger/MetricProducer_module.cc
eyandel: agreed. and again notes on where this work needs to happen are welcome!

lynnt-uchicago (approved):
generally looks good!! confirmed that my changes to the software trigger on the branch `/feature/lynnt_pmt_trig` build and look fine, up to you if you want to add the changes from that (mainly fcl parameter changes) to this commit or save for a future PR

sbndcode/Commissioning/HitDumper_module.cc
henrylay97: Fab, just checking :)

sbndcode/Commissioning/fcls/detsim_comm_fhicldump.out
henrylay97: ace

sbndcode/Commissioning/HitDumper_module.cc
henrylay97: Okay that's fair. Worth adding a comment above this line maybe to point us at a TODO for future.

sbndcode/Trigger/ArtdaqFragmentProducer_module.cc
henrylay97: Weird, okay!

sbndcode/Trigger/ArtdaqFragmentProducer_module.cc
```diff
@@ -0,0 +1,611 @@
...
+    }
+  } // trigger handle loop
+
+  if (fVerbose) std::cout << "Number of PMT hardware triggers found: " << triggerIndex.size() << std::endl;
+
+  // set properties of fragment that are common to event
+  uint32_t nChannelsTotal = channelList.size();
+  uint32_t nChannelsFrag = 15;
+  uint32_t nFrag = (uint32_t)nChannelsTotal/nChannelsFrag;
+  uint32_t wfm_length = 5120; // ~10us, 2ns tick
```
henrylay97, line 515: These (especially `wfm_length`) should not be hardcoded. Unlike the other example I think this is something we might want to change now Lynn Tung.

sbndcode/Trigger/ArtdaqFragmentProducer_module.cc
```diff
@@ -0,0 +1,611 @@
...
+      uint32_t fragmentIDVal = counter;
+      auto fragment_uptr = artdaq::Fragment::FragmentBytes(fragment_datasize_bytes, sequenceIDVal, fragmentIDVal, sbndaq::detail::FragmentType::CAENV1730, metadata); // unique pointer
+      fragment_uptr->setTimestamp(timestampVal);
+
+      // populate fragment header
+      auto header_ptr = reinterpret_cast<sbndaq::CAENV1730EventHeader*>(fragment_uptr->dataBeginBytes());
+
+      header_ptr->eventCounter = eventCounterVal;
+      header_ptr->boardID = boardIDVal;
+      header_ptr->triggerTimeTag = triggerTimeTagVal;  // ns // set timetag as random value for event
```
henrylay97, line 558: We need to fill the event size here. Its the only handle we have to read back the fragments from binary without assuming a set length of the waveform. As we're dealing with a lot of scary pointers and casting them I don't want to hardcode any sizes. We may also need other header field filled too but for now this one is the deal breaker.

henrylay97 (changes requested):
Hi both @lynnt-uchicago @eyandel 

In doing some work on the binary file stuff I have come across something that is important to this PR. I was getting seg faults when trying to properly serialize the waveform data from these simulated CAEN fragments. There is a field in the header called eventSize which is used to guide how big the waveform is that we're reading out. Could be hard coded but as I note in a comment that is a super bad idea when we're playing with pointers & casts so I really don't want to do that.

What we need to do is work out the eventSize in this module and fill this field so it can be used in the serialisation (but also for things like the EventAna or MultiDump modules, they both use this field and are also currently seg faulting when reading waveforms from these sim files).

sbndcode/Trigger/ArtdaqFragmentProducer_module.cc
henrylay97: I believe that the eventSize should be calculable as:

`wfm_length * (nChannelsFrag+1) / 2. + sizeof(CAENV1730EventHeader)`

The product `wfm_length * (nChannelsFrag+1)` gives you the number of waveform data points. We divide by two purely because the data points are saved as `uint16_t`'s and this field is a `uint32_t`. We then also must account for the header as it is stored before in the memory.

eyandel (commented):
should include all requested pmt changes from @henrylay97. Files made from this branch ran in the daq and gave sensible multidump output (and no seg faults seen). The hardcoded waveform length is now a fhicl parameter (which is what it is in the actual hardware as well).

henrylay97 (approved):
Looks great, thanks @eyandel!

fjnicolas:
trigger build SBNSoftware/sbnobj#62

fjnicolas:
@eyandel can you have a look at the c7 compiling error (https://dbweb8.fnal.gov:8443/LarCI/app/ns:SBND/build_detail/phase_details?build_id=sbnd_ci/10198&platform=Linux%203.10.0-1160.71.1.el7.x86_64&phase=build&buildtype=slf7%20c7:prof)?

sbndcode/Trigger/MetricProducer_module.cc
```diff
@@ -0,0 +1,509 @@
...
+  std::string fBaselineAlgo;
+  double fInputBaseline;
+  double fInputBaselineSigma;
+  int fADCThreshold;
+  bool fFindPulses;
+  double fPEArea; // conversion factor from ADCxns area to PE count
+
+  // event information
+  int fRun, fSubrun;
+  art::EventNumber_t fEvent;
```
henrylay97, line 84:
```suggestion
```

sbndcode/Trigger/MetricProducer_module.cc
```diff
@@ -0,0 +1,509 @@
...
+
+
+void sbndaq::MetricProducer::produce(art::Event& evt)
+{
+
+  // load event information
+  int fRun = evt.run();
+  art::EventNumber_t fEvent = evt.event();
+
+  if (fVerbose) std::cout << "Processing: Run " << fRun << ", Event " << fEvent << std::endl;
```
henrylay97, line 161:
```suggestion
  int run = evt.run();
  art::EventNumber_t event_num = evt.event();

  if (fVerbose) std::cout << "Processing: Run " << run << ", Event " << event_num << std::endl;
```

henrylay97 (commented):
The failure was an unused variable complaint which c7 is good at catching.

The variables were actually defined twice. I've suggested two commits which will remove the global instance, keep the local and rename the local to avoid them looking like global.

sbndcode/Trigger/MetricProducer_module.cc
eyandel: oh sorry I just saw this. I did the opposite (keep the global and delete the redefinition) , which I think should also resolve the issue. Is there a reason I should do your suggestion instead?

henrylay97: Absolutely fine, no particular reason no, I suppose as they're not needing to be accessed globally ever then that was my thinking. Yours works absolutely fine and other changes in that commit look fine too.

fjnicolas:
thanks @eyandel @henrylay97! It will be included in the next release

fjnicolas:
trigger build SBNSoftware/sbnobj#62

henrylay97:
If we could hold merging this for now, I think I might need a further edit to the CAEN sim fragment production. I'll update by the end of the day if not before, would be nice to save making a small extra PR!

sbndcode/Trigger/ArtdaqFragmentProducer_module.cc
```diff
@@ -0,0 +1,617 @@
...
+  // create and populate common metadata
+  sbndaq::CAENV1730FragmentMetadata metadata;
+  metadata.nChannels = nChannelsFrag + 1; // 15 PMT channels + final channel to store beam window / trigger information
+  metadata.nSamples = wfm_length;
+
+  // fragment handle properties
+  uint32_t eventCounterVal = fEvent;
+  uint32_t boardIDVal = 0;
+  uint32_t triggerTimeTagVal = (uint32_t)CLHEP::RandFlat::shoot(&fTriggerTimeEngine, 0, 1e9);
+  uint32_t eventSizeVal = ((wfm_length * (nChannelsFrag+1)) / 2.) + sizeof(sbndaq::CAENV1730EventHeader);
```
henrylay97, line 533: Aha, found the typo... We weren't dividing the event header size by 4.

```suggestion
  uint32_t eventSizeVal = ((wfm_length * (nChannelsFrag+1)) * sizeof(uint16_t) + sizeof(sbndaq::CAENV1730EventHeader)) / sizeof(uint32_t);
```

I don't know why but this field is stored in quad bytes. Not bytes. This is why we divide the length * channels chunk by 2 (each entry is 2 bytes) but we also need to do the same with the size of the event header, we want it in quad bytes, `sizeof` returns bytes.

I've changed the way we write it to try and make it more readable... Calculate the whole thing in bytes, then divide by the size of a quadbyte (4) but this is, in effect, the same as my suggestion the other day.

henrylay97 (changes requested):

sbndcode/Trigger/ArtdaqFragmentProducer_module.cc
henrylay97: @eyandel I will test this with the full system after lunch

henrylay97 (approved):
Sorry for the delay folks. This change now reflects the original intent and works through the full DAQ - Sim chain as it currently stands!

henrylay97:
trigger build SBNSoftware/sbnobj#62

PR #297, opened 2022-09-14
Persist PFO BDT features

henrylay97:
The sbnd equivalent edition to SBNSoftware/icaruscode#447

Changes the pandora xml to utilise newly added features in LArContent to persist extra metadata (PFO characterisation variables). CIwise it shouldn't change anything other than adding fields to PFOs in the CAFs.

### DIFF
--- a/sbndcode/SBNDPandora/scripts/PandoraSettings_Neutrino_SBND.xml
+++ b/sbndcode/SBNDPandora/scripts/PandoraSettings_Neutrino_SBND.xml
@@ -318,6 +318,7 @@
     <TrackPfoListName>TrackParticles3D</TrackPfoListName>
     <ShowerPfoListName>ShowerParticles3D</ShowerPfoListName>
     <UseThreeDInformation>true</UseThreeDInformation>
+    <PersistFeatures>true</PersistFeatures>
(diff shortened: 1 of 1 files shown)


fjnicolas:
@absolution1 are you able to review this PR so it can be included in the next release? Thanks!

henrylay97:
I believe Dom is now on holiday. @etyley could review in his place. It is a simple change and I have verified that the effects are as expected https://github.com/SBNSoftware/sbncode/pull/291#issuecomment-1248150598

etyley (commented):
Do we need to try and save the features twice? Are they saved separately or, as I suspect, does the second instance overwrite the first? If it is the latter I would advocate for removing the first instance if it serves no purpose.

henrylay97:
@brucehoward-physics and I discussed this. It definitely overrides the latter. My query was, if the PFOs are not amended by the intermediate algorithms do they have the process re-run?

etyley:
I am fairly sure that the second BDT will always run, as Pandora does not track whether the PFOs have been changed or not. Additionally, as the second BDT has a separate set of weights, even if it was unchanged the score would be different, so it should be run regardless.

henrylay97:
Okay, I'm happy to drop the initial instance then.

etyley (approved):

PR #298, opened 2022-09-15
Update Pandora cheating config for refactored larg4

etyley:


absolution1:
Thanks for this @etyley 
The changes look fine.  One quick Q: is this pre-plumbed into an already-existing fcl workflow?

etyley:
There is a fcl that does run the cheating but it was from before the fcl were refactored so is marked as `depreciated` so I did not bother updating it.

absolution1 (approved):
Thanks Ed.  We should think about undeprecating that other fcl, but that shouldn't stop this going through.
Approving now.

PR #304, opened 2022-10-10
Convert CI to use SCE enabled fcls

henrylay97:
Production have been using SCE as standard for quite a while. I thought we had updated the CI to reflect this quite a long time back. Checking my notes, we held off because we didn't have an operational sequential test to validate this with. 

This is a breaking change to the standard CI because the input files for the caf stage won't contain the SCE-type reconstruction products. I have run the [sequential](https://dbweb9.fnal.gov:8443/TestCI/app/ns:SBND/build_detail/phase_details?build_id=sbnd_ci/10387&platform=Linux%203.10.0-1160.71.1.el7.x86_64&phase=ci_tests&buildtype=slf7%20e20:prof) tests on this PR. I'm happy with the results and I think we should make this change as part of this week's release (not before testing other PRs) so we can reset before the next release.

Not entirely sure who should review so I've cast a little net... ;)


chilge:
makes perfect sense to do this.

chilge (approved):

wesketchum (approved):
Agree: great time to make this switch.

PR #305, opened 2022-10-18
Turn off blinding in cafmaker for MC

jacoblarkin:
Turn off newly added blinding code in cafmaker for MC since it's only needed for data (and SBND has no data cafmaker fcls right now). Added to cafmakerjob_sbnd.fcl which should propagate to other cafmaker fcls.


etworcester (approved):
Looks good

PR #314, opened 2023-01-10
software trigger metrics update and bug fix

lynnt-uchicago:
PR to update the software trigger metric producer modules. 

- more flexible configuration of the PMT software trigger producer. Added boolean fcl parameters to turn off and on various algorithms for lower/higher level trigger metrics. Updated in both the standalone PMT software trigger producer and the combined MetricProducer module 
- fixes bug where the start index and trigger index to create PMT artdaq fragments is incorrectly defined 
- fixes a bug where the software trigger data products would not be saved to the output root file. Fix involves producing the metric data products inside a `std::vector` (instead of a standalone data product). Fix applied to both the pmt software object and the CRT metric object. 
- Hitdumper module has been updated to reflect the above changes to read trigger data products 
- added an output root tree to the metric producer module and the PMT software producer module


lynnt-uchicago:
I have another an accompanying PR I need to create for small addition in sbnobj, but having trouble getting in to the gpvm's for some reason... will update when I'm able to make that PR!

lynnt-uchicago:
and question for @eyandel  and @mstancar , but do you think it makes sense to make separate folders within the Trigger folder for CRT and PMT standalone modules? the Trigger folder has a lot of files with similar names at this point, which may be confusing for newcomers. If so, I can update the CMakeFile and push in this PR.

eyandel:
> and question for @eyandel and @mstancar , but do you think it makes sense to make separate folders within the Trigger folder for CRT and PMT standalone modules? the Trigger folder has a lot of files with similar names at this point, which may be confusing for newcomers. If so, I can update the CMakeFile and push in this PR.

Yeah I think that would be nice. I don't think anything relies on these modules staying exactly where they are and the names are getting a bit confusing.

eyandel (approved):

mstancar (approved):

lynnt-uchicago:
ready to build!

PR #311, opened 2022-12-06
Accommodate breaking changes in geometry

knoepfel:
Pursuant to LArSoft/larcorealg#34.

PR #316, opened 2023-01-20
Bring the fast sim/reco unit tests into the 2020s

henrylay97:
We have two unit tests in SBND that run the sim/reco chain quickly on a single particle gun. These tests rarely fail so I have never paid to much attention to them. They did fail this week (whilst testing SBNSoftware/sbndcode#311) so I did... and I noticed:

- They are pretty out of date (using old fcls etc).
- They have slightly confusing names (the aim seems to be the same but with electrons and muons).

Thanks to @PetrilloAtWork I have done a little tinkering to rename them consistently and make use of newer, more appropriate fcls.

### DIFF
--- a/test/JobConfigurations/CMakeLists.txt
+++ b/test/JobConfigurations/CMakeLists.txt
@@ -1,6 +1,6 @@
 # for PREBUILT tests, the test name is the executable name, and no argument is allowed
-cet_test(prodsingle_reconstruction_test.sh PREBUILT)
-cet_test(fast_simulation_reconstruction_test.sh PREBUILT)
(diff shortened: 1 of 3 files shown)


PetrilloAtWork (approved):
It looks reasonable to me, but as I am not current on the standard workflow, I recommend that also @absolution1 check the changes.

PetrilloAtWork:
By the way, @newflux@, like "new coronavirus", expresses unrealistic expectations of the future...

henrylay97:
I'm happy that these are the standard workflow :+1:

fjnicolas:
trigger build

PR #337, opened 2023-04-26
Update CRT detector simulation

henrylay97:
Following extensive work to rewrite the CRT reconstruction for SBND the PR strategy has been agreed to be a series of PRs against a common branch (`feature/hlay_crt_clustering_base`) for the (inevitably very long) review stage. Please don't use CI tests yet. They will be used on the final PR when the common branch is compared to `develop`. At this stage the individual branches may not each compile on their own. The fully merged branch can be viewed here for completeness [feature/hlay_crt_clustering_merged](https://github.com/SBNSoftware/sbndcode/tree/feature/hlay_crt_clustering_merged).

This PR updates the detector simulation to:
a) Add more information contained in the real data (paired with SBNSoftware/sbnobj#79)
b) Remove the need for the CRTSlimmer (the new reco will start straight from FEBData)
c) Remove a process already handled in the geo wrapper and call such function


sbndcode/CRT/CRTSimulation/CMakeLists.txt
```diff
@@ -1,20 +1,23 @@
...
-                           cetlib::cetlib
-                           cetlib_except::cetlib_except
-                           messagefacility::MF_MessageLogger
-                           fhiclcpp::fhiclcpp
-                           ROOT::Core
-                           ROOT::MathMore
-                           sbnobj::SBND_CRT
-        )
+art_make_library(
+                 LIBRARY_NAME sbndcode_CRT_CRTSimulation
```
marcodeltutto, line 2: My understanding was that the library name is automatically generated to be in the format sbndcode_FOLDER_... and we should not explicitly type it here to avoid typos. Unless this has changed, I'd suggest removing this line.

henrylay97: Ahhh I did not know this! Tested offline and you're right. Have removed on this and other PRs :)

PR #341, opened 2023-04-26
crt clustering spacepointreco

henrylay97:
Following extensive work to rewrite the CRT reconstruction for SBND the PR strategy has been agreed to be a series of PRs against a common branch (`feature/hlay_crt_clustering_base`) for the (inevitably very long) review stage. Please don't use CI tests yet. They will be used on the final PR when the common branch is compared to `develop`. At this stage the individual branches may not each compile on their own. The fully merged branch can be viewed here for completeness [feature/hlay_crt_clustering_merged](https://github.com/SBNSoftware/sbndcode/tree/feature/hlay_crt_clustering_merged).

This PR introduces the characterisation of the CRTClusters, combining the information from the strip hits to give a single 3D position, time and PE value. The relevant object is created in SBNSoftware/sbnobj#84.


sbndcode/CRT/CRTReco/CRTClusterCharacterisationAlg.cc
```diff
@@ -0,0 +1,276 @@
+#include "CRTClusterCharacterisationAlg.h"
+
+namespace sbnd::crt {
+  
+  CRTClusterCharacterisationAlg::CRTClusterCharacterisationAlg(const fhicl::ParameterSet& pset)
+    : fCRTGeoAlg(pset.get<fhicl::ParameterSet>("GeoAlg", fhicl::ParameterSet()))
+    , fUseT1(pset.get<bool>("UseT1", true))
```
absolution1, line 7: Is the default value needed?  This can enable misconfigurations in fcl to go unnoticed

absolution1 (approved):
Looks good.
Approved, but I do have one question about a fcl default value.

sbndcode/CRT/CRTReco/CRTClusterCharacterisationAlg.cc
henrylay97: Probably not no, this is a field that would change for off beam running so yeah I'll move it into the fcl file and perhaps add a comment too.

PR #353, opened 2023-05-12
Feature/fnicolas pdssim

fjnicolas:
This PR includes various updates to the PDS simulation & reconstruction chain. A summary list is provided below. More details and validation plots can be found in [docdb-30445](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=30445).

- G4 stage: the ScintillationPreScale parameter is updated to 19%. This is done in order to match the new higher PMT detection efficiency.
- DetSim stage:
    - Updated PMT detection efficiencies: the PMT simulation now includes 3 detection efficiencies: two for the coated PMTs (VUV and VIS) and one for the uncoated PMTs.
    - Updated pedestal simulation: the baseline RMS has been updated from 1 to 2.6 ADC (according to the studies performed in [docdb-30879](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=30879)). The pedestal is now set to 15564 ADC. It assumes 95% signal-5% overshoot for the 14 bits CAEN V1730 digitizers.
    - Updates to the PMT saturation: we now simulate two saturation effects:
        - Saturation due to the digitizer dynamic range: truncates every waveform below 0 ADCs.
        - Includes PMT non linear effects by effectively re-scaling the #PE in each time bin.
- Reco1 stage:
    - Adds new method to estimate the OpFlash ZY barycenter
    - Updates how the PMTRatio metric (used to infer the OpFlash drift position) is calculated
    - Switch to a gauss filter for the PMT deconvolution

### DIFF
--- a/sbndcode/LArSoftConfigurations/opticalproperties_sbnd.fcl
+++ b/sbndcode/LArSoftConfigurations/opticalproperties_sbnd.fcl
@@ -27,8 +27,10 @@ sbnd_opticalproperties: {
     
     ScintByParticleType: true
 
-    ScintPreScale:  0.03   # QE x TPB_FW x TPB_Absorption (15% x 50% x 45%)
-                           # MUST match between g4 and detsim
+    # ScintPreScale MUST be equal/larger than the largest detection efficiency applied at DetSim stage
+    # This corresponds to the uncoated PMTs detection efficiency (9.6%)
+    # See sbndcode/OpDetSim/digi_pmt_sbnd.fcl
+    ScintPreScale:  0.19
 
     EnableCerenkovLight: false # Cerenkov light OFF by default
 
--- a/sbndcode/LArSoftConfigurations/opticalproperties_sbnd.fcl
+++ b/sbndcode/LArSoftConfigurations/opticalproperties_sbnd.fcl
@@ -28,7 +28,7 @@ sbnd_opticalproperties: {
     ScintByParticleType: true
 
     # ScintPreScale MUST be equal/larger than the largest detection efficiency applied at DetSim stage
-    # This corresponds to the uncoated PMTs detection efficiency (9.6%)
+    # This corresponds to the uncoated PMTs detection efficiency (19%)
     # See sbndcode/OpDetSim/digi_pmt_sbnd.fcl
     ScintPreScale:  0.19
 
--- a/sbndcode/OpDetReco/OpDeconvolution/job/sbnd_flashfinder_deco.fcl
+++ b/sbndcode/OpDetReco/OpDeconvolution/job/sbnd_flashfinder_deco.fcl
@@ -5,7 +5,7 @@ BEGIN_PROLOG
 ####OpFlash finder for PMT deconvolved waveforms#####
 ###TPC0
 SBNDDecoSimpleFlashTPC0: @local::SBNDSimpleFlashTPC0
-SBNDDecoSimpleFlashTPC0.PECalib.SPEAreaGain: 500
+SBNDDecoSimpleFlashTPC0.PECalib.SPEAreaGain: 200
 SBNDDecoSimpleFlashTPC0.OpHitProducers: ["ophitpmt"]
 SBNDDecoSimpleFlashTPC0.OpHitInputTime: "RiseTime"
 SBNDDecoSimpleFlashTPC0.UseT0Tool: true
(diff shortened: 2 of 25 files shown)


sbndcode/LArSoftConfigurations/opticalproperties_sbnd.fcl
rodralva, line 1: make sure this doesn't underestimate  the smearing of the truth number of photons arriving at the PMTs/XARAPUCAs

sbndcode/OpDetReco/OpDeconvolution/job/sbnd_flashfinder_deco.fcl
rodralva, line 1: Note for @rodralva (me), the drift coordinate estimation is only applied to the PMTs for now, need to add the calibration curve before adding the XARAPUCAs here

sbndcode/OpDetReco/OpFlash/FlashTools/DriftEstimatorPMTRatio_tool.cc
rodralva, line 1: same comment, will either have to create separate tool for XARAPUCAs or modify current tool to take both subsystems into account

sbndcode/OpDetReco/OpFlash/FlashTools/FlashGeoBarycenter_tool.cc
rodralva, line 1: Small picky comment here, I would call this 2d barycenter, or ZY barycenter or something similar. As we are also estimating the X coordinate and someone could think the tool is doing the complete 3d XYZ vertexing here

sbndcode/OpDetSim/CMakeLists.txt
rodralva, line 1: Non used code removed

rodralva (approved):
Immense amount of work. I don't like the fact that we have similar repeated functions form workers up to the digiAlg for both PMTs and XAs, but that's another topic to discuss.

Side note: I would stress the fact that the barycenter tool only computes the Y and Z coordinate, and not the X (which is done in the drift tool)

fjnicolas:
trigger build

fjnicolas:
trigger build

fjnicolas:
trigger build

fjnicolas:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_75_00 SBNSoftware/sbncode@v09_75_00

fjnicolas:
Comments on the CI results:
- Fhicl-dump outputs: all differences are expected from the changes introduces by this PR
- G4 stage: 
  - Increase in the size of the `sim::OpDetBacktrackerRecord` products for the light generated in the outer volume: expected since the ScintPreScale has been increased by a factor of ~6 (more chances that a photon generated in the out volume gets to on of the photon detectors and hence creates an instance of the OpDetBacktrackerRecord)

```
1792: < G4 | pdfastsimout | Reflected | std::vector<sim::OpDetBacktrackerRecord> | 223
1793: ---
1794: > G4 | pdfastsimout | Reflected | std::vector<sim::OpDetBacktrackerRecord> | 294
```

- DetSim stage: reduction in the number of `OpDetWaveforms`. This is in principle unexpected. This PR increases the detection efficiencies for the PMTs, so we may expect more optical detectors triggering/above threshold (ie more waveforms).

```
669: < DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 4429
670: ---
671: > DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 3545
672: 98c98
673: < DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 5523
674: ---
675: > DetSim | opdaq |  | std::vector<raw::OpDetWaveform> | 4258
```

The reason we are seeing less waveforms is that we also modified the `ScintPreScale` parameter (from 0.03 to 0.19) in order to match the higher detection efficiency set in the DetSim modules (19% for uncoated PMTs). The efficiency corrections applied in the DetSim module are given by (XARAPUCA case, same logic applies to the PMTs)

```
 , fXArapucaVUVEff(fParams.XArapucaVUVEff / fParams.larProp->ScintPreScale())
 , fXArapucaVISEff(fParams.XArapucaVISEff / fParams.larProp->ScintPreScale())
```

The efficiency corrections applied at DetSim stage with and withouth this PR are:

| | wo PR        | wPR           | 
| ------------- |:-------------:| -----:|
| Scint PreScale | 0.03 | 0.03 |
| XARAPUCA VUVEff| 0.021      | 0.021  | 
| XARAPUCA VIsEff | 0.014      |  0.014  |
| XARAPUCA VUVEff/ScintPreScale| 0.7      | 0.11  | 
| XARAPUCA VIsEff/ScintPreScale | 0.46     |  0.07  |

This means that in practice, efficiency correction factor applied by the digitizer modules is smaller with this PR.  The reference G4 input sample was done using the old 0.03 `ScintPreScale`, resulting in less number of waveforms above threshold. This is not axtually an issue and it should be back to normal once the reference samples are updated to use the new `ScintPreScale` value.

- Reco1 stage: there is a drop in the number of `OpHits` for the PMTs

```
874: < Reco1 | ophitpmt |  | std::vector<recob::OpHit> | 46420
875: ---
876: > Reco1 | ophitpmt |  | std::vector<recob::OpHit> | 42236
```

With this PR, PMTs are using a Gausss filter (instead of a Wiener filter). The Gauss filter results in wider OpHits, meaning that some OpHits that are close in time are merged into a single OpHit. An example is shown below. This results in the overall reduction in the number of reconstructed OpHits.

     - Wiener: 
Wiener
     - Gauss: 
Gauss

PR #365, opened 2023-07-07
Lynn's updates to save waveforms from software metrics

bear-is-asleep:
Lynn fixed some bugs where the beam was not being found in the software metric. She also added a fcl option to save the raw waveforms. I did some proof checking with my NC light filter studies and it works well! I've been given her blessing to make a PR

### DIFF
--- a/sbndcode/Trigger/PMT/pmtSoftwareTriggerProducer_module.cc
+++ b/sbndcode/Trigger/PMT/pmtSoftwareTriggerProducer_module.cc
@@ -122,7 +122,18 @@ private:
   bool   _beam_trig; 
   double _time_trig; 
   int    _npmt;
-  double _promptPE, _prelimPE; 
+  double _promptPE, _prelimPE;
+
+  TTree* _pulse_tree;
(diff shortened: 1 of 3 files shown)


eyandel (approved):
This looks fine for now but we should discuss making some things (e.g. baseline, pull window lengths) configurable. I think the best thing would be to compare and converge with the daq version of this code. But that can wait until the daq version has real pmt signals to be tested with.

RachelCoackley:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_75_02 SBNSoftware/sbncode@v09_75_02 SBNSoftware/sbnanaobj@v09_21_02 SBNSoftware/sbnobj@v09_17_03

lynnt-uchicago:
@eyandel The baseline is fcl configurable, I believe the pull window might also be? We specify a waveform length (that matches the artdaq producer module for simulation), but certainly there are other things that would probably be nice as a fcl param.

lynnt-uchicago (approved):

PR #385, opened 2023-09-07
Fix off-by-one bug in channel map call in TPC decoder

tomjunk:
Apologies for noisy PR's -- I see the channel ID in the Nevis map starts at 0 and not 1, so this is the bugfix.  I also made a separate fcl file with a PROLOG in it for decoder configs that can be #included by other fcls more easily.


nupuroza (approved):

kalradaisy (approved):

fjnicolas:
trigger build

PR #387, opened 2023-09-21
call removeProduct for raw fragments to conserve memory

tomjunk:
The raw fragments, once decoded, do not need to linger in the art event.  art provides removeProduct for this purpose, but it only works for products that have been read from a file rather than produced in the job, as they can be re-read from the file if need be.

I had to change the daq_handle from a validHandle to just a Handle because removing the product makes the handle invalid, so removeProduct() is not even a method of validHandle.

The memory savings are not obvious in a job that just runs the decoder, as there is nothing else to use the memory until the next event.  But in a job that produces other things, it should reduce the memory footprint by the size of the raw TPC fragments.

### DIFF
--- a/sbndcode/Decoders/TPC/SBNDTPCDecoder_module.cc
+++ b/sbndcode/Decoders/TPC/SBNDTPCDecoder_module.cc
@@ -100,8 +100,11 @@ daq::SBNDTPCDecoder::Config::Config(fhicl::ParameterSet const & param) {
 
 void daq::SBNDTPCDecoder::produce(art::Event & event)
 {
-  auto const& daq_handle = event.getValidHandle<artdaq::Fragments>(_tag);
-
+  auto daq_handle = event.getHandle<artdaq::Fragments>(_tag);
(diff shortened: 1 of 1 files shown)


kalradaisy (approved):

fjnicolas:
trigger build

tomjunk:
Weird -- the changes I made are not yet in any tested code.  We probably should include a test for this.  In any case,
I found that the commit here was swept into another PR I put in today:

https://github.com/SBNSoftware/sbndcode/pull/397

for the PTB decoder.

tomjunk:
Anyhow, the crashed thread in the CI tests is in the famous libunwind.so.  Do you see this in other CI test runs?

fjnicolas:
Hi @tomjunk, yes we are seeing the crash for `c14` with all the other PRs (also the rest of experiments are also having the same issue). The warnings for `e26` are all understood and unrelated to this PR (we have just updated our ref files after the transition to the new compilers). I'm planning to merge this PR in the next release if you agree.

PR #411, opened 2024-02-09
Adding new SCE map for the 2d drift simulation

jzennamo:
The 2D drift simulation uses the wrong convention for the SCE offsets in the right hand TPC.

This PR adds a new set of fcls which uses as "hacked" map to correct this issue in the 2D drift simulation. 

The reconstruction and the 1D drift simulation should use the "standard" SCE maps. 

...I am sorry for having to make this PR....

It also requires an update to `sbnd_data`:

`/exp/sbnd/app/users/jaz8600/SpaceChargeFix/srcs/sbnd_data`

### DIFF
--- a/sbndcode/LArSoftConfigurations/enable_spacecharge_services_sbnd_2D_DriftSimOnly.fcl
+++ b/sbndcode/LArSoftConfigurations/enable_spacecharge_services_sbnd_2D_DriftSimOnly.fcl
@@ -1,0 +1,11 @@
+# Author: Gray Putnam <grayputnam@uchicago.edu>
+# Include this file at the bottom of any fhicl file where space-charge is intended
+# to be turned "ON"
+# It will modify all services parameters necessary.
+
+services.SpaceCharge.EnableCalSpatialSCE: false
+services.SpaceCharge.EnableSimSpatialSCE: true
+services.SpaceCharge.EnableSimEfieldSCE: true
+services.SpaceCharge.EnableCalEfieldSCE: false
+
+services.SpaceCharge.InputFilename: "SCEoffsets/SCEoffsets_SBND_E500_voxelTH3_2DSigSimHack.root"
--- a/sbndcode/JobConfigurations/standard/detsim/2d_drift_sim/CMakeLists.txt
+++ b/sbndcode/JobConfigurations/standard/detsim/2d_drift_sim/CMakeLists.txt
@@ -1,0 +1,1 @@
+install_fhicl()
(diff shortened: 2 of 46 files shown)


fjnicolas:
@jzennamo I just pushed an updated version of `sbnd_data` (still `v01_23_00`, so no need to update the `product_deps`)

sbndcode/LArSoftConfigurations/enable_spacecharge_services_sbnd_2D_DriftSimOnly.fcl
```diff
@@ -0,0 +1,11 @@
+# Author: Gray Putnam <grayputnam@uchicago.edu>
+# Include this file at the bottom of any fhicl file where space-charge is intended
+# to be turned "ON"
+# It will modify all services parameters necessary.
```
henrylay97, line 4: Maybe a comment here to record your hack for posterity @jzennamo? Might help when we unpick this in future ;P

henrylay97 (commented):
I approve in general of this approach. Will hold off approving until the necessary validation has been done offline to show this definitely works (and doesn't requires any reco-level adjustments as well).

In the meantime I have a comment about the fcls. I fully appreciate the approach being taken but until a proper refactor of the fcls is achieved most SBND fcls _aren't_ running 2D sim. Therefore most of these new fcls won't actually do what they claim to do.

My recommendation would be to remove all these extra fcls, and just edit:
- `detsim_sce_lite_wc.fcl`
- `g4_sce_dirt_filter_lite_wc.fcl`
To have:
`#include "enable_spacecharge_services_sbnd_2D_DriftSimOnly.fcl"` at the bottom.

Definitely not pretending that this situation is ideal but I think it keeps it controllable and reasonably high in the fcl chain such that the inevitable new fcls for the upcoming production should follow the same template.

lynnt20:
helloooo! Looking at this PR, perhaps it'll make sense in the (hopefully) near future to merge it with my `feature/lynnt_wirecell_cfgs` branch once that branch has been reviewed a bit. That should clean up having to use specialized names for what _should_ be a standard 2d workflow.

jzennamo:
*Update on PR:*

Instead of calling new maps I have added some code changes to `SpaceChargeSBND.cxx` along with a fcl flag. 

Lane is doing a reco level validation now but here is my sim level validation:
image (4)

fjnicolas:
@jzennamo can we drop the hacked SCE map from `sbnd_data` then?

jzennamo:
Lane has validated that these changes work! 

refac_test_offsets

fjnicolas:
As agreed offline, this PR is closed as the relevant commits to fix the SCE bug are introduced in #416.

PR #413, opened 2024-02-12
Quick fix to the hitfinder module to work with the Raw XARAPUCA waveforms

rodralva:
Fixed:
-Polarity was set to negative as PMTs
-Thresholds where too high (current SER peak is 8 ADC)
-Computed the Area of the current SER (no idea what was implemented before, 3 years since the last change)

Changed:
-Slightly tuned the algorithm params.


fjnicolas (commented):
@rodralva you said you had to change the pulse polarity? I don't see that updated in the fhicl.

The new paramters look sensible to me at first sight, though some more tuning for the workflow with the raw waveforms may be needed (same applies to the PMTs).

rodralva:
@fjnicolas you are absolutely correct. I forgot to add it from the custom original fcl that I prepared. Also fixed the area computation (I loaded the wrong SER file).

fjnicolas (approved):

PR #409, opened 2024-02-01
SimChannel Module Label

henrylay97:
This changes the module label used by the back tracker for SimChannels from `simdrift` (1D simulation) to `simtpc2d:simpleSC` (WireCell 2D simulation).

Details are in [docDB #33131](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=33131).

Like with the interplane drift effect this invalidates the use of these fcls on 1D simulation files but I think given the lack of parallel workflows this is the best approach.

I change the base use of this in the definition of the backtracker service _plus_ other module label locations that are currently pointed at `simdrift`. @lynnt20 feel free to weigh in with a WireCell perspective.

Issue #407, opened 2024-01-26
CAF stage cannot be run with c14 builds


fjnicolas:
Running the CAF stage with `c14` qualifiers will result in

`Can't find file "sbn_cnnid.fcl"`

from `sbndcode` `v09_82_02`.

We are currently running the [CNN hit classiffication module](https://github.com/SBNSoftware/sbncode/tree/develop/sbncode/TPCReco/CNNHitClassification) as part of the CAF stage (see #386). This module depends on TensorFlow, for which LArSoft currently only provides `e26` builds.

Note: by now we are only providing `e26` builds for the SBN/SBND managed repositories, so it's not actually an issue right now.
I'm just opening it for future reference if people attempt to run CAF stage once `c14` builds are back.

bear-is-asleep:
Won't fix since c14 build is not used anyways. Will fix should the necessity arise

PR #414, opened 2024-02-16
Update unit tests to use WireCell workflow

fjnicolas:
Intended to match #409

### DIFF
--- a/sbndcode/JobConfigurations/standard/caf/cafmakerjob_sbnd.fcl
+++ b/sbndcode/JobConfigurations/standard/caf/cafmakerjob_sbnd.fcl
@@ -2,6 +2,7 @@
 
 #include "services_sbnd.fcl"
 #include "geometry_sbnd.fcl"
+#include "simulationservices_sbnd.fcl"
 
 #include "larproperties.fcl"
 #include "backtrackerservice.fcl"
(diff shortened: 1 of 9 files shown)


fjnicolas:
trigger build

lynnt20:
Looks like the CI is still calling the old version of `standard_g4_sbnd.fcl` that includes `mcreco`. I think this may be expected?

lynnt20:
Found another instance of `mcreco` in a g4 stage, in `g4_dirt_filter.fcl`. Unsure if this is the cause for the unit test failure, but should probably be changed in this PR 🥲  

https://github.com/SBNSoftware/sbndcode/blob/0f5b7c3ae8bd25509f3d624626203da94b97c3a5/sbndcode/JobConfigurations/standard/g4/g4_dirt_filter.fcl#L11

henrylay97 (approved):
Yeah looks good to me - once it's been shown Lynn's commit solves the failure.

@lynnt20 I notice some of the wirecell fcls have quite a lot of "other detsim things" commented out? Not necessarily one for this PR but might be worth removing to avoid confusion?

fjnicolas:
> Looks like the CI is still calling the old version of `standard_g4_sbnd.fcl` that includes `mcreco`. I think this may be expected?

@lynnt20 what test do you mean? This PR only fixes the unit tests (they look fine to me in the previous build, other than the known issue with `c14` and CAF maker #407). The fix for the CI is not addressed here (coming PR from Rachel).

lynnt20:
@fjnicolas oh I see, I misunderstood!! Everything looks good to me then!!

lynnt20 (approved):

PR #379, opened 2023-07-25
Feature/apapadop sbnd systtools

afropapp13:
feature branch containing the ffcl's for implementation of systtools in the sbnd workflow

PR #417, opened 2024-02-23
Systtools Implementation for SBND

henrylay97:
This PR is a replacement for PR #379 now that @jedori0228 has migrated the systtools common configs to sbncode (SBNSoftware/sbncode#433).

It purely replaces instances of `genieweight` (the old implementation of GENIE ReWeight using SBNEventWeight) with `systtools`.

I also renamed the fcls to reflect this change (so they say what they are doing on the tin & follow the same naming convention as the ICARUS fcls).

I removed an instance of `genieweight` from a fcl that claims to only run `fluxweight`.

And finally, I point the relevant CI fcl at the updated fcl. Unlike with SBNEventWeight we cannot override the number of universes in the CI fcl. With @jedori0228 's blessing I have added a `test` configuration to the systtools directory in sbncode which reduces the multisims to just 5 universes. This is used in the CI & can be used by developers during testing.

I have done a small validation using `cafmakerjob_sbnd_sce_systtools_and_fluxwgt.fcl`  and established that it runs fine and places relevant weights in the CAF file.


henrylay97:
Responding to the CI tests.

I've updated the fcl checks list to reflect the fcl name change. Other than that the differences look expected / unrelated. Hard to parse the CAF checks due to the large scale of upstream changes _but_ the weights do change and change in a way that looks sensible to me!

fjnicolas:
@sjgardiner @afropapp13 @jedori0228 the parent PR was just merged, would you mind reviewing/approving? thanks!

afropapp13 (commented):
Looks good to me! if there are no further comments from anybody else, i would be happy to approve the changes

fjnicolas (approved):
PR approved in comments by @afropapp13

PR #399, opened 2023-12-19
add fcls for mlreco bnb sample

bear-is-asleep:
Producer and g4 fcls for bnb + cosmics sample. Dependent on larg4 changes
* https://github.com/LArSoft/larg4/pull/52 to store dropped particles as separate vector
* https://github.com/LArSoft/larg4/pull/51 to store orig track ID

PR #416, opened 2024-02-20
Using WireCell in standard fcl workflow by default

lynnt20:
This PR will fully integrate the WireCell 2D TPC Simulation/Signal Processing chain into SBND's "standard" fcl workflow. With this updated workflow, there is no longer a separate WireCell fcl/stage; the WireCell stage is merged into the standard detsim stage. This PR follows PR #409 , PR #408, and PR #414. 

Note that the drift simulation, TPC electronics simulation, and deconvolution all happen within the _detsim_ stage now. Previously the 1d drift sim (simdrift) was run in g4, the electronics sim (daq) was run in detsim, and 1d deconvolution (caldata) was run in reco1. 

A fix for space charge effects with 2D sim is now implemented by @jzennamo, where the last commit from PR #411 is cherry-picked into this PR. The 2D "fix" space charge service should be called on all modules upstream of/including detsim that use sce. 

`sbndcode/JobConfigurations/README.md` has been updated with the most recent "core" fcl workflow.

# Table of contents
1. [Standard/common fcls](#standard)
2. [2D SCE Fix/fcls](#sce)
3. [Detvar fcls](#detvar)
4. [WireCell fcls](#wirecell)
5. [Commissioning fcls](#commissioning)

## Changes to standard/common fcls <a name="standard"></a>
- `standard_g4_sbnd.fcl`: simdrift no longer run! mcreco moved to end of detsim
  - WireCell requires that ionandscint products are *not* dropped in the g4 stage. 
  - `g4_dirt_filter.fcl` (the standard g4 fcl for rockbox) no longer runs simdrift
  - `g4_simphotontime_filter.fcl` (the standard g4 fcl for in-time cosmic) has all references to simdrift/mcreco removed

- `standard_detsim_sbnd.fcl`: runs WireCell 2D Sim/SigProc
  - will output `sim::SimChannel` and `recob::Wire` by default. Configuration allowing`raw::RawDigit`s to be saved available in the commissioning fcl folder (see below).
  -  note: the electron drift _simulation_ is in the _detsim_ stage, instead of g4. 
  - note: mcreco is run at the end of detsim, since it requires `SimChannels`
 
- `reco_sbnd.fcl`: caldata no longer run! (1D deconvolution now obsolete)
  - gaushit and caloskim (calibration ntuple module) have labels updated to WireCell input
  - fasthit is now disabled in reco1, this is because we are not saving `raw::RawDigit`s by default. Additionally, no other downstream modules were using fasthit output. 

[...]

PR #418, opened 2024-02-23
ML Reconstruction chain becomes friends with Wirecell

bear-is-asleep:
Updated ML Reco fcls from this [PR](https://github.com/SBNSoftware/sbndcode/pull/399) to standardize the workflow, and merge it with the wirecell workflow.


bear-is-asleep:
> @bear-is-asleep does this PR completely recplaces #399 (once the WireCell fhicl refactoring gets in)?

PR #399 will not work with #416 changes. We could get one tagged release out before #416 is merged, but otherwise #399 will be outdated after those changes. I'm working with @lynnt20 to make the ML workflow compatible with the WC changes, which is the goal of this PR.

bear-is-asleep:
Too many things leaked in and I suck at github so I've move the PR

PR #423, opened 2024-02-29
Offline PMT Decoder

lynnt20:
Offline PMT Decoder that reads in `artdaq::Fragment`s from the CAEN 1730 Digitizers and outputs `raw::OpDetWaveform`s.  This PR is not necessary for the production release. 

Main features: 
- stores a timestamp for the `raw::OpDetWaveform` based on other timing subsystems, such as SPEC TDC and PTB. In the case that the products for the SPEC TDC or the PTB do not exist, the timestamp that will be stored is the start of the waveform w.r.t. the CAEN timing reference system.
- combines waveforms from extended triggers. An extension is identified if two conditions are followed: (1) the length of the waveform is shorter than the nominal (set by fcl param) and (2) if the start of the short waveform is less than 10 us away from the previous waveform.
- outputs histograms of the waveforms for a single event (event number changed w/ fcl param). 
- this decoder does not perform jitter correction; the timing corrections will be performed in another downstream module.  

There are more features that I'd like to add to the decoder, but wanted at least a working version of it available on `sbndcode` with the hope that we'll soon be looking at real PMT waveforms 😄. Such features include: 
- FTRIG thresholds specified per board (instead of one FTRIG threshold for every board)
- storing FTRIG tick into a data product rather than saving the entire waveform into artroot (which currently can only turn on/off the option to save)


tomjunk (approved):
Looks good!  We still have some old getByLabel calls in there instead of newer things, but the art team has not made any more noises about deprecating it.  As long as it's tested, we should get this into a release.

fjnicolas:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_85_00

lynnt20:
found a bug, need to do one more commit before merge!

fjnicolas:
> found a bug, need to do one more commit before merge!

OK! btw, all the CI warnings are obviously unrelated

lynnt20:
ok should be good now :)

PR #426, opened 2024-03-11
TPC channel map inverting the order of FEMCh numbers in groups of 8

tomjunk:
Tingjun and Varuna ran some channel-ID mode data on March 9.  Tingjun found that the channel numbers reported by the channel-ID mode data did not match those in the channel map -- they had an inverted ordering in groups of eight.  This would have gone largely unnoticed in our current studies of noise vs channel number because it rearranges channels in nearby groups of eight.  But we did see some mysterious behavior with blips seen in run 10880 that looked like blips were split into pairs way more often than expected.  I rearranged the channel map so the FEMCh numbers now count the other way in groups of eight relative to the initial try at the channel map.

Since this is a bugfix, I am not increasing the version number of the channel map text file, merely replacing it; there is no reason to keep the old one simultaneously in the release.  I will post valdation plots in this PR.


tomjunk:
Here is the FEMB channel number observed vs. expected using the old v1 channel map.  Plot from Tingjun
fembchannelnumber

tomjunk:
And here are buggy and fixed event displays from the same event with a blip in it from Run 10880
evd twq-proj 10880 1899_tpc1
evd_invertedfemch

tomjunk:
Here is the fixed FEMB Channel number plot from Tingjun with the new map
fembchannelnumber(1)

nupuroza (approved):

PR #429, opened 2024-03-14
Add Signal Processing only fcl + configuration

lynnt20:
Adds a fcl configuration for running the signal-processing stage _only_; current fcls run both tpc simulation + sigproc. The default input label for the signal-processing fcl is `daq`, which is the label for data (not simulation). Note that the output label for `recob::Wire` is different: `sptpc2d` instead of `simtpc2d`. In the future, we may want to change all labels to something like `tpc2d`.

`wcsimsp_sbnd.fcl` has been updated to provide a skeleton configuration (`sbnd_wcls`), and new configurations `sbnd_wcls_simsp` (sim+sigproc) and `sbnd_wcls_sp` (sigproc only) inherit from this skeleton configuration. In the future, additional configuration blocks can be added into this single fcl. 

I have run `fhicl-dump` on the new version and previous version of `wirecell_simsp_sbnd.fcl` to confirm there are no real changes to the applied configuration, and have also run test jobs (over 2 events) to make sure that sim+sigproc run as expected. The signal-processing only fcl (`wirecell_sp_sbnd.fcl`) also runs over real decoded noise data. 

Note that noise filtering is **off** for all current jsonnets/fcls.


ebelchio12 (approved):
Thanks Lynn, looks good. This is a good time to push this fcl and update current jsonnets

fjnicolas:
trigger build SBNSoftware/sbndutil#13 SBNSoftware/sbndaq-artdaq-core@feature/mnebot_art3.14.04 SBNSoftware/sbn*@release/SBN2024A

PR #412, opened 2024-02-11
New XAs PDEs to VUV and visible light

rodralva:
New data-driven proposed PDEs based on CIEMAT measures and analysis, summary can be found here:
[https://docs.google.com/presentation/d/17mvsNqaZfgkvTXj8s4t61ZNk1JXJdCT6hNuyQc7cJvg/edit?usp=sharing](
https://docs.google.com/presentation/d/17mvsNqaZfgkvTXj8s4t61ZNk1JXJdCT6hNuyQc7cJvg/edit?usp=sharing)

PR #435, opened 2024-03-20
Include APA mesh atteenuation factor in the PMT efficiencies

fjnicolas:
Updating the Photon Detection Efficiencies after the new measurements for the X-ARAPUCAs (#412) we realized that the light attenuation due to the APA grounding mesh was not accounted for the PMTs. This component is not present in the GDML file for the SBND geometry (i.e. not accounted by the semi-analytic model corrections), so it needs to be included in the overal detection efficiency. This PR is to include its effect in the simulation of the PMT signals.

The mesh transparency is 85% [[ref](https://sbn-docdb.fnal.gov/cgi-bin/sso/RetrieveFile?docid=15081&filename=grounding_mesh_status_update_102319.pdf&version=1)], but the light attenuation depends on the photons incident angle, as can be seen in the figure below

The distribution of incident angles for the two components VUV and visible are shown below

We see that the maximum of the distribution is at 42º for the VUV and at 37º for the visible light. A zoom of the attenuation factor around those values can be seen below

According to these results we will use in our simulations an attenuation factor of 0.8


sbndcode/OpT0Finder/job/opt0finder_sbnd.fcl
```diff
@@ -59,8 +59,9 @@ sbnd_opt0_finder_one_to_many.FlashVetoTimeEnd: 2
...
... 0.,0., 0., 0., 0.021, 0.021, 0.021, 0.021, 0.021, 0.021,0.021, 0.021, 0.021, 0.021, 0.021, 0.021, 0., 0., 0.,0., 0., 0., 0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.096 ,0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.021, 0.021, 0.021,0.021, 0.021, 0.021, 0., 0., 0., 0., 0., 0.,0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.096 ,0.096 , 0.096 , 0.096 , 0.021, 0.021, 0.021, 0.021, 0.021, 0.021,0., 0., 0., 0., 0., 0., 0., 0., 0.,0., 0., 0., 0., 0., 0., 0., 0., 0.,0.021, 0.021, 0.021, 0.021, 0.021, 0.021, 0.021, 0.021, 0.021,0.021, 0.021, 0.021, 0., 0., 0., 0., 0., 0.,0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.096 ,0.096 , 0.096 , 0.096 , 0.021, 0.021, 0.021, 0.021, 0.021, 0.021,0., 0., 0., 0., 0., 0., 0.096 , 0.096 , 0.096 ,0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.096 ,0.021, 0.021, 0.021, 0.021, 0.021, 0.021, 0., 0., 0.,0., 0., 0., 0., 0., 0., 0., 0., 0.,0., 0., 0., 0., 0., 0., 0.021, 0.021, 0.021,0.021, 0.021, 0.021, 0.021, 0.021, 0.021, 0.021, 0.021, 0.021,0., 0., 0., 0., 0., 0., 0.096 , 0.096 , 0.096 ,0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.096 , 0.096 ,0.021, 0.021, 0.021, 0.021, 0.021, 0.021]
```
lynnt20, line 62: thank you for going through all of these and changing them 🥰

lynnt20 (approved):

henrylay97 (approved):
Nice one Fran, thanks for all the validation. As discussed offline I'm happy with this PR going in. The only direct impact on CRUMBS will be through the flash measured PE variable from OpT0Finder, but this is not a crazy change. Doesn't require an immediate update, can just be rolled into the next training automatically :D

PR #437, opened 2024-03-20
Update PMT baseline

bear-is-asleep:
PMT ADC baselines are now at 14k not 8k, so we need to update the baseline values.


fjnicolas:
trigger build

PR #439, opened 2024-03-20
TPC Decoder invalid fragments -> warning instead of exception

tomjunk:
Tingjun requested that the behavior of the TPC decoder be changed from throwing an exception (this was the old behavior with the call to getValidHandle, but retained with the getHandle call), to a warning.  The decoding is skipped and an empty raw digit collection stored in the event when the input handle is invalid.  This enables DQM jobs to continue past invalid events.


fjnicolas:
the new (production) tag will be coming on Monday; @yangtj207 are you happy with the proposed changes?

yangtj207 (approved):

PR #440, opened 2024-03-20
evd_sbnd_data.fcl with drawing options for raw data drawing

tomjunk:
My custom fcl file for making raw data event displays -- just a few convenience options.


fjnicolas:
@yangtj207 happy with this fhicl?

yangtj207 (approved):

PR #441, opened 2024-03-23
Add new functions in geometry helper

yangtj207:
Add a function to calculate the intersection point of two wires. 
Add an analyzer module to analyzer channel ID data.


tomjunk (approved):
I'll go ahead and approve this, but the person doing the merge should check PR 443 as well -- there are commits with the same comments and what looks like the same changes in that.  Maybe the same branch was used for both PRs?

fjnicolas:
> I'll go ahead and approve this, but the person doing the merge should check PR 443 as well -- there are commits with the same comments and what looks like the same changes in that. Maybe the same branch was used for both PRs?

thanks for the warning @tomjunk, looks like that's the case. I'll take it into account for the merging process.

fjnicolas:
trigger build

PR #446, opened 2024-03-26
Update dirt flux path

VCLanNguyen:
Validation of dirt flux is in [sbn-docdb#33374](https://sbn-docdb.fnal.gov/cgi-bin/sso/RetrieveFile?docid=33374&filename=SBND%20Flux%20Update.pdf&version=1)
Wiki entry is updated [here](https://sbnsoftware.github.io/sbndcode_wiki/The_SBND_flux_files.html)

**New updates:**

- Files were produced using BooNEtoGSimple on GPVM instead of the legacy MiniBooNE machines. 
- Kaon re-weight is applied with a stand-alone code and does not have MiniBooNE dependency.

**Comparing between the new (config L) vs current dirt flux (config I) :**

- Fluctuations is < 3% for muon neutrino and antineutrino.
- Saw a discrepancy ~5% at energy for electron neutrino and ~10% for electron antineutrino. In particular, the peak at 300 MeV in the neutrino energy spectrum is slightly larger in this new flux (config L). 
- The current dirt flux  (config I) was compared against the standard TPC flux generated at the same time (config H), and it was found to be consistent. 
- After checking with Zarko, this confirmed that both the current dirt flux (config I) and the standard TPC flux generated at the same time (config H) did not have kaon re-weighting when generated.
- Therefore, the fluctuation in the new dirt flux (config L) is due to including kaon re-weighting.
-  The current standard TPC flux (config K) already includes the kaon re-weighting.


VCLanNguyen:
trigger build

marcodeltutto (approved):
Looks good, thank you Lan!

PR #448, opened 2024-03-28
Offline PMT Decoder Updates

lynnt20:
Updates to the offline PMT decoder:

1. Change the internal organization structure of the vector of fragments to store _by board_ instead of _by trigger_. Because the decoder is not performing jitter correction, there is no need to index by trigger and compare the trigger time between boards for the same trigger. The internal re-structure to store the vector of fragments by board will still allow us to combine waveforms for extended trigger. This restructure also allows us to easily open up `CAENV1730` fragments, whereas before we were mainly opening up `ContainerCAENV1730`. 
2. Update channel mapping to version in docdb 33668-v9.
3. Add a fcl parameter to allow a fragment ID mapping to the board ID for old DAQ configurations. This can also be utilized in the future if the fragment ID->board ID mapping changes. 

These changes have been tested in recent PMT data runs (such as the PMT ramp up gain runs) as well as older PMT runs to ensure that there exist fcl configurations to run over all types of files.


tomjunk:
HI Lynn -- looks like a bugfix just went in about an hour ago!1  Looks like good work.  Thanks for renaming the hardcoded map from the hack map -- hacks have a way of sticking around for a long time and to call them hacks does not give the user much confidence.  In fact, it's not even hardcoded -- it's read from fcl.  ProtoDUNE-SP actually hardcoded their map in the compiled code (ouch).  Probably not worth renaming it however.

The std::find and std::distance may be better off as a std::unordered_map lookup.  But I'll bet the CPU usage of looking up stuff in this map is close to negligible anyway.

Where is the jitter correction to be applied?  Do we need a new collection of jitter-corrected OpDetWaveforms made by a downstream module?

What do you do if the fragid is not in the map provided in fcl?  Looks like the code just falls through without updating fragid, even if fuse_hardcoded_map is true.  fignore_fragid might be better as a std::set, where the lookup is O(logn) in CPU.

tomjunk:
Thanks!  I never understood the need for container fragments anyway.  Got enough containers containing the data.  But that's how the data come up?

lynnt20:
haha good question, this is my understanding of what is container vs. non-container fragments: 

- if a CAEN board is being used to _push_, then the fragments from that board will appear as normal fragments. All other boards that receive the trigger (pull mode) will have their fragments appear in a container. If all boards are in pull mode, all fragments appear in a container. 
- the number of entries in a container is equal to the number of triggers

So I think that container vs. non-container is a somewhat useful distinction :) but does require some additional work to use.

tomjunk:
Okay, sounds good.  If you're done with this PR, I can approve it.  I assume you've been testing it with real DAQ data?  It's okay to modify it more in the future, this just releases it for everyone else.

tomjunk (approved):
Looks good!

lynnt20:
trigger build

lynnt20:
trigger build

tomjunk:
What's the status here?  I think the CI failures were independent -- some issue with the AutoLoadingVisitor:

50: Error in cling::AutoLoadingVisitor::InsertIntoAutoLoadingState:
51:    Missing FileEntry for FieldResponse.h
52:    requested to autoload type larcv::FieldResponse
53: Error in cling::AutoLoadingVisitor::InsertIntoAutoLoadingState:
54:    Missing FileEntry for SegWeightInstance3D.h
55:    requested to autoload type larcv::SegWeightInstance3D
56: Error in cling::AutoLoadingVisitor::InsertIntoAutoLoadingState:
57:    Missing FileEntry for XYZ2PixelCoordinate.h
58:    requested to autoload type larcv::XYZ2PixelCoordinate
59: Error in cling::AutoLoadingVisitor::InsertIntoAutoLoadingState:

A similar set of messages was seen in the event display:

https://cdcvs.fnal.gov/redmine/issues/28702

The comment in LArSoft v09_89_01's release notes:

https://github.com/LArSoft/larsoft/releases/tag/v09_89_01

is Revert to using deprecated cet_rootcint()

I saw from some of the CI logs above that AutoLoadingVisitor errors also come from supera.

PR #452, opened 2024-04-04
Compute asicchan from FEMBCh and fill the field in ChanInfo

tomjunk:
I had put in a field called "asicchan" in the ChanInfo struct but forgot to fill it.   It can be computed from FEMBCh.  asic is just FEMBCh / 16.  No one has complained yet, but it's best not to leave the field uninitialized.


yangtj207 (approved):

yangtj207:
Looks good. Approved.

PR #457, opened 2024-04-11
update trigger to use 14.7k baseline

bear-is-asleep:
PMT baseline is now at 14.7k baseline instead of 8k. Update the trigger to reflect this. 

** Disclaimer: I have not tuned these sufficiently. They've just been corrected to use updated baseline.**


fjnicolas:
@eyandel @lynnt20 @pgreen135 can you have a look?

eyandel (approved):

PR #463, opened 2024-04-30
increase precision of timestamp calculation to 64 bits -- was 32 bits

tomjunk:
The TPC Analysis data product that the raw decoder makes had a 32-bit timestamp which is inadequate.  I had copied it from here:

https://github.com/SBNSoftware/sbndqm/blob/bbdb305238b32b4d3d215bba47ea1bf44ba97b6c/sbndqm/Decode/TPC/HeaderData.hh#L21

I changed the data type to ULong64_t from uint32_t in the data product, and upgraded the formula computing it so that it doesn't roll over at 2E9 or so.  The classes_def.xml file automatically updated, and ROOT's automatic schema evolution should take care of reading in old files as only the size of one element of the data product changed.  I checked reading in an old and a new analysis data product in the same ROOT session in the same TTree, and ROOT figured it out.  Below are results of a test of redecoding a file and looking at the timestamp vs. the entry number in the file.

run12804_rollover
norollover


kalradaisy (approved):

fjnicolas:
@tomjunk I'll include this in the next release

PR #466, opened 2024-05-03
Trj digitalnoisechannelstatus may3 2024

tomjunk:
This is a service that checks raw::RawDigit and/or recob::Wire to identify channels that have digital noise on them.  The selection is on the RMS with a fcl-steerable parameter (separate for raw digits and recob::Wire), or if the differences between raw digit values is always divisible by 2, which covers cases where we've seen that bits are shifted and differences are powers of two.  Looks like I re-used a branch for my old timestamp PR that should have been merged already.


tomjunk:
I haven't tried the recob::Wire test and the default module label may need to be updated -- currently it's "caldata" for that.  It also flags most channels in sunset events as "bad" due to large RMS.

yangtj207 (approved):

fjnicolas:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_89_01 SBNSoftware/sbncode@v09_89_01

PR #474, opened 2024-05-29
Add fcls tuned to pmt data

bear-is-asleep:
The hitdumper can run successfully with pmt decoded data. However, some tuning is needed for the OpHitFinder for the raw waveform, which is included in a separate fcl. I've also added a separate configuration fcl, that will only read data. For now it only reads pmts, but as more subsytems come on they can be added. 

image
Above is all pmt waveforms from run 12877. The red lines show the expected region of noise `PedRangeMax/Min` and the blue line is the ADC threshold, which is the same as the standard fcl, 13 ADC.


sbndcode/Commissioning/fcls/hitdumpermodule_data.fcl
```diff
@@ -0,0 +1,47 @@
+BEGIN_PROLOG
+
+hitdumper:
```
lynnt20, line 3: i'd rename this to `hitdumper_data` or something to keep things consistent...

lynnt20: actually, the best thing to do is probably create a new configuration that inherits from the default, rather than an entirely new one that has most things copy-and-pasted... let me find an example

lynnt20:
instead of making an entirely new config fcl that is nearly identical, I would suggest adding onto the original config fcl file and call another instance with a new name, that has some fcl parameters changed. Example [here](https://github.com/SBNSoftware/sbndcode/blob/develop/sbndcode/WireCell/wcsimsp_sbnd.fcl) (wirecell configs).

The base config is `sbnd_wcls`, and `sbnd_wcls_simsp` and `sbnd_wcls_sp` inherit from the base one. I think this will ultimately lead to less confusion down the line, especially if you change the defaults for the base, then it'll propagate through all other configs that reference it.

sbndcode/Commissioning/fcls/hitdumpermodule.fcl
```diff
@@ -27,7 +27,7 @@ hitdumper:
  
     keepCRThits:              true
     keepCRTstrips:            false
-    makeCRTtracks:            true
+    makeCRTtracks:            false
```
bear-is-asleep, line 30: I added this since we won't be making CRT tracks in hitdumper anymore and it's bugged anyways.

sbndcode/Commissioning/fcls/hitdumpermodule_data.fcl
bear-is-asleep: Done, this was a great suggestion, thanks!

marcodeltutto (approved):
Thanks, @bear-is-asleep!

bear-is-asleep:
trigger build

PR #477, opened 2024-06-11
As built CRT geometry, v1

djpayne-thing:
Updates the CRT geometry to something a bit closer to what we built. Does not include gaps, etc, but this is more or less what exists. Does not include the skinny modules under the cryo valve. Top CRT is closer to what will actually exist than it was. It's a process.


bear-is-asleep:
Hi @djpayne-thing , can you assign a reviewer please?

bear-is-asleep:
It's typical to have someone else review, so there are multiple eyes on it. Are there any other CRT experts who can check the geometry? @henrylay97 comes to mind.

fjnicolas:
> Looks like I can assign myself. Is that appropriate?

@djpayne-thing it must be someone different than the PR author (actually GitHub doesn't allow you to do that). Probably simulation conveners or people previously working on geometry (e.g. @gustavogx @mrguzzo) are a good options

djpayne-thing:
gdml1
gdml2
This is what the changes I've made look llike

marcodeltutto (commented):
Thanks, @djpayne-thing! Should we also update `geometry_sbnd.fcl` so this new geometry will be picked up?

djpayne-thing:
Added a separate pull request for geometry_sbnd.fcl

PR #479, opened 2024-07-05
Setting cafmaker.SaveGENIEEventRecord to true for MC CAFMaker

jedori0228:
With https://github.com/SBNSoftware/sbncode/pull/459, we now have ability to save GENIE event record to CAFs. This feature is turned on when having `SaveGENIEEventRecord` to `true` in the CAFMaker parameter. In icaruscode, we opened a PR to have this parameter true for default MC CAFMaker: https://github.com/SBNSoftware/icaruscode/pull/739
I'm trying to do the same in sbndcode here; but as a NON sbnd-expert, I am not sure if this is the correct fhicl to change, but looks like to be the one? Someone from SBND should chime in here and correct me if I'm wrong.. Also, do we have data version of this in SBND? We should make sure there we have this parameter `false`.
More details can be found in here: https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=36869


lynnt20:
Thank you @jedori0228 !!!

bear-is-asleep:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_90_00

henrylay97 (approved):
Not sure whether my review adds anything now but this looks good, thanks @jedori0228.

We don't yet have a data CAFMaker fcl but I imagine that is something we will have very soon so all of us should remember to ensure we make the required change.

bear-is-asleep:
Approved and passes the working CI.

PR #478, opened 2024-06-13
To pick up geometry v02_01

djpayne-thing:

PR #480, opened 2024-07-09
CRT Channel Map for GDML v02 01

henrylay97:
This is the channel map update to pair with PR #478

It provides the following
- The mapping for the extra flat module created in the GDML update
- The correct inversions for the newly shortened modules
- A change to the positions and inversions for all vertical modules in the east wall which Alex & David found to be incorrect.
- A change in the logic in the CRTGeoAlg due to the swapping of the two layers in the north wall.

I appreciate Tom Junk's point about numbering in my previous PR. I am hoping that v6 will actually turn into productions v1, I will think of a better numbering scheme at that point. We need to be slightly more certain that it is right first though.


djpayne-thing (approved):
As shown in [https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=34844](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=34844 ) these changes give us the expected module orientation.

bear-is-asleep:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_90_00 SBNSoftware/sbncode@v09_90_00

PR #486, opened 2024-07-25
Add frame number, sample number, and artdaq fragment timestamp to TPCDecodeAna data product

tomjunk:
This PR adds three new members to the TPCDecodeAna data product -- the frame number which comes up in the Nevis TPC fragment for each FEM, the 2 MHz sample number, and the artdaq fragment timestamp.  The first two numbers are added together to make the Nevis timestamp in microseconds which already was in the TPCDecodeAna data product, but which we want to inspect separately.  The artdaq fragment timestamp is put in by the board reader upon receipt of the data, and helps us debug the choppy events we've been seeing.


kalradaisy (approved):

bear-is-asleep:
trigger build

PR #498, opened 2024-08-25
CRT GDML Fixes

henrylay97:
## Description 
This PR fixes two outstanding issues relating to the CRT GDML
- The inversion of the east & west walls.
- The mis-assignment of strips to modules in the bottom CRT following GDML v02_01

It also makes a couple of minor aesthetic changes visible in the diffs from each commit.

There are other CRT GDML updates that are somewhat less pressing that will be followed in a future PR. See [docDB #37693](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=37693) for details of the changes made and those needed in future.

Note that, as module numbers do not change this PR does _not_ require a follow up channel map change.

## Checklist
- [x] Added at least 1 label from [available labels](https://github.com/SBNSoftware/sbndcode/issues/labels?sort=name-asc).
- [x] Assigned at least 1 reviewer under `Reviewers`,
- [x] Assigned all contributers including yourself under `Assignees`
- [ ] Linked any relevant issues under `Developement`
- [ ] Does this PR affect CAF data format? If so, please assign a CAF maintainer as additional reviewer.
- [ ] Does this affect the standard workflow? 

### Relevant PR links (optional)
None

### Link(s) to docdb describing changes (optional)
See above

PR #496, opened 2024-08-23
Feature/lkashur geo wires2

lkashur:
## Description 
Update no. of collection plane wires in simulation to 1670.   This is a slight change to the gdml (and base gdml), see sbnd_v02_02.gdml.  Then mask the central 6 wires to account for the gap between APAs, and one wire on each half APA to account for wire combs, bringing us to the correct number of collection plane wires (1662).  These changes are incorporated in sbndcode/Utilities/badchannels_sbnd.fcl and sbndcode/Utilities/channelstatus_sbnd.fcl.

maskedcolwires

PR #497, opened 2024-08-24
Updated configurations for Wire-Cell noise filtering

ebelchio12:
## Description 
Updated configurations for Wire-Cell noise filtering. The updates in this PR are the new overall response used for signal protection in coherent noise filtering (consistent with SBND field response and electronics response we are currently using. Previous was consistent with ProtoDUNE one), and separate NF configurations for data/MC. There is also a new bad channel masking (updated list of bad channels) and RC response deconvolution in MC (included in noise filtering step). The changes are all about signal processing stage but for data it is separated in reco1 stage, while for MC it is unified with detsim stage.

1. To have WireCell reflect the changes in the wire gdml, we need to update sbnd-wires-geometry-v0*.json.bz2 in the **sbnd_data** directory. The new geometry file can be found here: /exp/sbnd/data/users/lynnt/wirecell/geometry/new_geom/sbnd-wires-geometry-v0202.json.bz2

2. It should go inside the WireCell directory in the corresponding **sbnd_data** version.

PR #487, opened 2024-07-30
Bad channels from data, and a data-specific sbnd_services block

absolution1:
This PR implements the list of bad/dead/noisy channels from docdb: [37116](https://sbn-docdb.fnal.gov/cgi-bin/sso/RetrieveFile?docid=37116&filename=BadChannels_20240717.pdf&version=1)

Key features:
 - A set of bad channel sequences organised by type
 - A dedicated channel service fcl block for data
 - A dedicated table of data-specific services (this block is currently just the channel service)
 - Inclusion of the data-specific service block in `reco1_data.fcl`, `reco2_data.fcl` and `evd_sbnd_data.fcl`

Important items to note: 
 - I've opted to complete omit the list of bad channels from MC into the list of bad channels for data, because of the issue with the total number of channels not being correct.
 - ~~This PR is marked as draft because my testing has so far been limited to a fcl dump to confirm the fcl mechanically runs.  Opening up an event display is still TODO~~

PR #499, opened 2024-08-26
Feature/lkashur crt colwires gdml fix

lkashur:
## Description 
This PR updates SBND's simulated geometry, including changes to the CRT (#498) and TPC collection plane wires (#496).  Also update are the "bad channels" list (sbndcode/Utilities/badchannels_sbnd.fcl and sbndcode/Utilities/channelstatus_sbnd.fcl) to go along with the collection plane wires fix. 

Dependent on all PRs in project [34](https://github.com/orgs/SBNSoftware/projects/34) to be ready, since fcl tables are affected by new channel mapping IDs.

## Checklist
- [ ] Does this PR affect CAF data format? If so, please assign a CAF maintainer as additional reviewer.


lkashur:
Just ended up making a new branch to include both our changes @henrylay97.  Could you verify that everything CRT-related made it in here?

bear-is-asleep:
Is this going to change any channel numbers that need to be considered in PRs like #497 and #487?

henrylay97:
Yep, this looks good to me! Thanks @lkashur!

henrylay97 (approved):

marcodeltutto:
Thank you again, @lkashur! And also for merging the CRT work!

@tomjunk @yangtj207 This PR changes the number of collection plane wires. I am tagging you again as this replaces PR #496. 😄 Could you please take a look?
As Lane wrote above, there are now 1670 wires on the collection planes. When saving data, the 6 wires in the middle should be skipped.
This PR keeps the number of induction wires to 1984 but removes all the "bad" channels in these planes so that we can use all the 1984 channels for both simulation and data.

@absolution1 The bad channels from data (PR https://github.com/SBNSoftware/sbndcode/pull/487) should be updated. Is this something you could help with?

tomjunk:
Yes, will take a look, though Tingjun and I are on vacation this week as Fermilab requires it.

tomjunk:
Dom is also out until the end of August.

tomjunk:
I checked the wire dumps -- so far the number of channels per plane looks as expected.  I recomputed the bad channel numbers and they agree with Haiwang's update of the bad channel list on the last page of SBN-doc-37116-v2.

tomjunk:
I updated the channel map text file (now sbndcode/ChannelMaps/TPC/SBNDTPCChannelMap_v2.txt, keeping the old one in case someone wants to use the old GDML) and updated the fcl config to point to it.  I redecoded a file from run 15040 using the new channel map and ran the larsoft event display with evd_sbnd_data.fcl.  I searched for an event which has a steep track that crosses the APA gap in Y on the west and made an event display, see below.

n.b.  With the other commits in this PR, the default GDML one is now the v2.2 which is great.  I ran the MC gen and sim (though I had to use the legacy detsim without CRT as WireCell seemed to not initialize when I tried it), and noticed each time the GDML was loaded, the V2.2 one was chosen, and that was true with the event display too.  This is great, but reading in an old decoded data file with the new geometry will show errors in the event display or reco.  The TPC decoder does not actually use the geometry, and so existing decoded files, such as those in the commissioning area in dCachek, will not trigger LArSoft's check on consistency of the geometry.  They will still read in, but the channel numbering will not be right for the new geometry.  Instead, one ought to re-decode the raw data with the new channel map.
evd twq-proj 15040 319

tomjunk:
I also updated SBN-doc-35006-v1 with a new cartoon and wire end dump text file and root file.

tomjunk:
It looks like the decoded files in the commissioning area in dCache indeed do have the geometry information in their configuration.  So while they still need to be re-decoded in order to use the data with the v2.2 geometry, attempting to read in one that was decoded with the v2.0 geometry into a job using the v2.2 geometry ought to generate an exception from the geometry configuration writer.

tomjunk:
A separate problem I had was the legacy detsim fcl couldn't find crtslimmer_sbnd.fcl

lar -c legacy_detsim_sbnd.fcl g4.root
Failed to parse the configuration file 'legacy_detsim_sbnd.fcl' with exception
---- search_path BEGIN
  Can't find file "crtslimmer_sbnd.fcl"
---- search_path END

Art has completed and will exit with status 90.

tomjunk:
It would be good to see if (U, V, Y) hit triplets match better now, now that the Y hit locations are ±9 mm displaced.

marcodeltutto:
@tomjunk can you share the path to your redecoded file from run 15040?

bear-is-asleep:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_91_02

bear-is-asleep:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_91_02

bear-is-asleep:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_91_02

bear-is-asleep:
Hi @tomjunk @marcodeltutto @yangtj207 @dombarker30 , can one of you review/approve the PR? Thanks!

marcodeltutto:
I would like to have the following approvals on this PR:
- @tomjunk for the decoder part 
- either @yangtj207 or @dombarker30 for the reconstruction side

tomjunk (approved):
Yes, I approve for the decoder/channel map and the new TPC channel numbering and locations in the v2.2 GDML file.

tomjunk:
And I can answer my own question about whether points match up better in the three views with the new geometry, and indeed they do.  Here is the same track, which crosses the APA gap, in the old geometry and old channel map, and again with the new geometry and new channel map.  The magenta stars come from the LArSoft event display's shift-click 3D space point finder, with the two input points being Induction 1 and Induction 2 (U and V), and the predicted point is in collection.  There is a 3-wire offset, as expected, and the new geometry is clearly a better match.  The track may actually end partway through the argon volume observed by any given wire, so the points still don't match exactly.

trackend_oldgeo
starpts_trackend_newgeo

tomjunk:
And here's a randomly-selected track endpoint (I think the track goes through the APA), in the old and new geo.  The 3 wire spacings help a lot.
example2_newgeo
example2_oldgeo

yangtj207 (approved):
Looks good to me.

marcodeltutto (approved):
Thanks!

PR #520, opened 2024-09-16
bulletproof the call to removeProduct -- deal with case if handle is not valid

tomjunk:
Just a little bulletproofing for the call to removeProduct().  I saw in the shift operations Slack channel some mention about an invalid handle from the TPC decoder running in the DQM.  I looked through the source to see if there was a test on isValid() before doing anything, and to gracefully handle cases in which the data are missing.  It was all okay except the call to removeProduct() was not guarded.  It may be that the call will still not throw an exception even if the handle is not valid, but I figured I'd put in a test anyway.

This is a standalone PR.  I tested it by redecoding the same file I redecoded for PR499 and reproduced the event displays, not re-uploaded here.


bear-is-asleep:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_91_02

yangtj207 (approved):
Looks good.

PR #535, opened 2024-10-14
update timesize (the size of the frame in samples minus one) to 1143

tomjunk:
This PR updates the timesize variable, which is the size of the FEM frame in samples minus 1, per a communication with Georgia and Seokju.

The timesize variable is used in calculating the single 64-bit timestampls put in raw::RDTimeStamp data prducts for the TPC, and also for the tpcAnalysis::TPCDecodeAna data products:

https://github.com/SBNSoftware/sbndqm/blob/bbdb305238b32b4d3d215bba47ea1bf44ba97b6c/sbndqm/Decode/TPC/SBND/SBNDTPCDecoder_module.cc#L60

The tpcAnalysis::TPCDecodeAna data product contains the frame number and the sample number so the timestamp can be recomputed should this number change (it's a DAQ configuration parameter and the timestamp is recomputed offline).

Here is a copy of the conversation I had in Slack with Seokju and Georgia:

-------------------------------------------------------------------------------------------------------

Tom Junk, October 10:

Hmm.  I cannot remember changing timesize.  In develop, it is still 2559.  At least we have both the frame number and the sample number in the TPCAnalsyis data product if we need to recompute timestamps from earlier data, as we are about to embark on a processing of the data.  But this value should be set to something that makes sense for new data.  What is the right timesize for new data?  Thanks!

Seokju Chung, October 11:

  [1:10 PM](https://shortbaseline.slack.com/archives/C07FKMP4TM4/p1728670203948319)
Since Daisy is away, I'll try to check with Georgia or Nupur to get the correct value

  [1:54 PM](https://shortbaseline.slack.com/archives/C07FKMP4TM4/p1728672883143779)
This is Georgia's reply
The frame size configuration parameter in the hardware should be set to the value of 9151. This corresponds to a frame size which is (9151+1)/8=1144 2MHz samples. The readout size configuration parameter in the hardware should be set to the value of 1143. This (I expect) should correspond to a readout window of (3x1144) +/-1=3432 +/-1 2MHz samples (if running without huffman compression), and possibly as low as 12-14 fewer 2MHz samples (after uncompressing, if running wihtout huffman compression).
From a private conversation
| Oct 9th
[1:55](https://shortbaseline.slack.com/archives/C07FKMP4TM4/p1728672952021149)
So, the timesize is 1143 if you’re asking for the daq config parameter, and 1144 if you’re trying to calculate how many samples are in each event and each channel (3x1144) +/-1=3432 +/-1 2MHz samples, uncompressed

------------------------------------------------------------------------------------------------------------

No other PRs need to be merged with this one, either in sbndcode or any other repository.

It affects the production workflow by changing the number used in the calculation of RDTimeStamps.  The choppy-event filter should work even if the frame size is not right -- it only checks to see if things are different.  This update is needed if we are to synchronize the TPC with the rest of the detector.


tomjunk:
If I see Michelle, I can ask her to approve.  See also slide 7 of SBN-doc-37831-v1 for documentation on this new number.  There's a nomenclature issue of whether timesize has the -1 in it already or not, but as the number has 1 added to it in the decoder to make the timestamp, 1143 appears to be the right value for this fcl parameter.

mstancar (approved):

bear-is-asleep:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_93_01 SBNSoftware/sbnanaobj@bug/coackley_ciCafv09_91_02

bear-is-asleep:
Ready for merge

PR #542, opened 2024-10-22
fcl to run all the decoders

tbwester:
## Description 
Add a fcl file to run all the decoders on raw data files. This fcl is the first stage of keep-up processing. This PR is a draft while there is on-going decoder development.

## Checklist
- [x] Assigned all contributors including yourself under `Assignees`
- [ ] Linked any relevant issues under `Development`
- [ ] Does this PR affect CAF data format? If so, please assign a CAF maintainer ([PetrilloAtWork](https://github.com/PetrilloAtWork) or [JosiePaton](https://github.com/JosiePaton)) as additional reviewer.


tomjunk:
Very good, thanks for this!   I redecoded an already-decoded file from run 15040 and found sensible TPC, PTB and NTB data.  The PTB data look better-decoded than the keepup file due to bugfixes.  I see it is still a draft PR, so I won't sign off on it just yet.

sbndcode/Decoders/run_decoders_job.fcl
```diff
@@ -0,0 +1,108 @@
...
+    TFileService: {
+        fileName: "decoder_hist.root"
+    }
+}
+
+physics:
+{
+  producers:
+  {
+      daq: @local::SBNDTPCDecoderDefaults
```
lynnt20, line 54: @tomjunk I was wondering if it might it make sense to change the reference to the TPC decoder from `daq` to something more consistent like `tpcdecoder`? 

I believe the `daq` label is historical from the 1D TPC simulation, and in fact our current 2D TPC simulation uses a different producer label entirely. I don't mean to suggest that @tbwester needs to update this PR, but it might be a good change to make in a future PR with the blessing of each of the relevant folks? I assume there will be a number of downstream changes. 

At first even I was a little confused why I didn't see `tpcdecoder` in the nice list of decoders before realizing that it was labelled `daq` 😛

henrylay97 (approved):
Happy from the CRT side :)

sbndcode/Decoders/run_decoders_job.fcl
tomjunk: Yes, I noticed this too. I think `daq` is historical, and MC (!) driven.  I am okay with any name, and `tpcdecoder` sounds more fitting with the pattern.  But the dowstream reco fcls would have to be adjusted too.

bear-is-asleep:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_93_01 SBNSoftware/sbnanaobj@bug/coackley_ciCafv09_91_02

bear-is-asleep:
Ready to merge

PR #541, opened 2024-10-21
PMT Decoder Updates: Timing Data Product + Timing CAEN addition + more

lynnt20:
## Description 

1. Creates a two new data products to persist CAEN/event timing information for the downstream in case we drop the `CAEN1730` Fragments. 
   - `raw::pmt::eventTimingInfo`: saves which timing system and what channel/word was used for the event reference time.
   - `raw::pmt::boardTimingInfo`: saves the postpercent and a vector of TTT (size will be >1 if there are extended triggers) 
   - Association are also made between `raw::OpDetWaveforms` and `raw::pmt::boardTimingInfo`. Because there are three different instances of `OpDetWaveform`s created in each event, there are also three different instances of the associations. 
2. Adds logic for choosing the PTB HLT timestamp when the SPEC TDC ETRIG is missing 
   -  the HLT with the smallest trigger word >0 and <10000 (these limits are fcl configurable) is chosen, a check is also done to ensure the timestamps are within 3 ms (fcl configurable) of the raw header timestamp 
3. adds the timing CAEN to the default decoding configuration (AKA the timing CAEN will be decoded by default)
   - timing CAEN waveforms do not need special treatment for decoding
   - timing CAEN waveforms from ch0, ch1, ch2 will be saved under instance `TimingChannels` as `raw::OpDetWaveforms`, with channel numbers `900`,`901`,`902` respectively. 
   - timing CAEN waveforms from ch15 will be saved into the pre-existing instance `FTrigChannels` as `raw::OpDetWaveforms`
   - added fcl configurables to turn on/off saving the above waveforms, and which channels to save
4. adds some flexibility depending on number of TDC ETT found in the decoded TDC objects
   - if more than 1 TDC ETT is found, then the one closest to the raw header timestamp will be used as the reference time 
   - correction for raw header timestamp is done by default, value set by `raw_ts_correction`, default value is 367 us.

## Checklist
- [x] Added at least 1 label from [available labels](https://github.com/SBNSoftware/sbndcode/issues/labels?sort=name-asc).
- [x] Assigned at least 1 reviewer under `Reviewers`,
- [x] Assigned all contributers including yourself under `Assignees`
- [N/A] Linked any relevant issues under `Developement`
- [N/A] Does this PR affect CAF data format? If so, please assign a CAF maintainer ([PetrilloAtWork](https://github.com/PetrilloAtWork) or [JosiePaton](https://github.com/JosiePaton)) as additional reviewer.
- [x] Does this affect the standard workflow?

PR #570, opened 2024-12-16
Feature/lynnt pmt decoder

bear-is-asleep:
## Description 

PR #541

1. Creates a two new data products to persist CAEN/event timing information for the downstream in case we drop the `CAEN1730` Fragments. 
   - `raw::pmt::eventTimingInfo`: saves which timing system and what channel/word was used for the event reference time.
   - `raw::pmt::boardTimingInfo`: saves the postpercent and a vector of TTT (size will be >1 if there are extended triggers) 
   - Association are also made between `raw::OpDetWaveforms` and `raw::pmt::boardTimingInfo`. Because there are three different instances of `OpDetWaveform`s created in each event, there are also three different instances of the associations. 
2. Adds logic for choosing the PTB HLT timestamp when the SPEC TDC ETRIG is missing 
   -  the HLT with the smallest trigger word >0 and <10000 (these limits are fcl configurable) is chosen, a check is also done to ensure the timestamps are within 3 ms (fcl configurable) of the raw header timestamp 
3. adds the timing CAEN to the default decoding configuration (AKA the timing CAEN will be decoded by default)
   - timing CAEN waveforms do not need special treatment for decoding
   - timing CAEN waveforms from ch0, ch1, ch2 will be saved under instance `TimingChannels` as `raw::OpDetWaveforms`, with channel numbers `900`,`901`,`902` respectively. 
   - timing CAEN waveforms from ch15 will be saved into the pre-existing instance `FTrigChannels` as `raw::OpDetWaveforms`
   - added fcl configurables to turn on/off saving the above waveforms, and which channels to save
4. adds some flexibility depending on number of TDC ETT found in the decoded TDC objects
   - if more than 1 TDC ETT is found, then the one closest to the raw header timestamp will be used as the reference time 
   - correction for raw header timestamp is done by default, value set by `raw_ts_correction`, default value is 367 us.

## Checklist
- [N/A] Linked any relevant issues under `Developement`
- [N/A] Does this PR affect CAF data format? If so, please assign a CAF maintainer ([PetrilloAtWork](https://github.com/PetrilloAtWork) or [JosiePaton](https://github.com/JosiePaton)) as additional reviewer.
- [x] Does this affect the standard workflow?


bear-is-asleep:
Already approved and passed unit tests. #541

PR #581, opened 2024-12-17
Feature/maxd ptb decoder update

maxdubnowski:
## Description 
Description added to the document included

## Checklist
- [ ] Added at least 1 label from [available labels](https://github.com/SBNSoftware/sbndcode/issues/labels?sort=name-asc).
- [ ] Assigned at least 1 reviewer under `Reviewers`,
- [ ] Assigned all contributers including yourself under `Assignees`
- [ ] Linked any relevant issues under `Developement`
- [ ] Does this PR affect CAF data format? If so, please assign a CAF maintainer ([PetrilloAtWork](https://github.com/PetrilloAtWork) or [JosiePaton](https://github.com/JosiePaton)) as additional reviewer.
- [ ] Does this affect the standard workflow? 

[PTB Decoder Update (1).pdf](https://github.com/user-attachments/files/18159471/PTB.Decoder.Update.1.pdf)

PR #586, opened 2024-12-18
Revert "Feature/maxd ptb decoder update"

bear-is-asleep:
Reverts SBNSoftware/sbndcode#581. Accidentally merged to develop. I believe we can allow this to stay in develop but we can revert in case this isn't ok.


bear-is-asleep:
Reverts https://github.com/SBNSoftware/sbndcode/pull/581. Accidentally merged to develop. Will repopen once `run_ptbana.fcl` is fixed.

RachelCoackley (approved):

bear-is-asleep:
This is not necessary anymore, we will just do a bug fix to the fcl. It's a single line that needs to be removed

PR #593, opened 2025-01-13
Feature/maxd post581 p rfix

maxdubnowski:
## Description 
There were minor adjustments to PR 581 that didn't make it in. It requires at least sbndaq_artdaq_core v1_10_03 to run.

### DIFF
--- a/sbndcode/Decoders/PTB/SBNDPTBDecoder_module.cc
+++ b/sbndcode/Decoders/PTB/SBNDPTBDecoder_module.cc
@@ -203,9 +203,11 @@ void SBNDPTBDecoder::_process_PTB_AUX(const artdaq::Fragment& frag, ptbsv_t &sou
 		tstruct.gate_counter = ctbfrag.Trigger(iword)->gate_counter;
               }
               else if (*(frag.metadata<int>()) != 2){ //If data is taken with 128b PTB words
-		tstruct.prev_timestamp = 0;
-		tstruct.gate_counter = 0;
+		tstruct.prev_timestamp = -1;
+		tstruct.gate_counter = -1;
 	      }
+
+	      std::cout << tstruct.prev_timestamp << "   ||   " << tstruct.gate_counter << std::endl;
 
 	      ix = sout.HLTrigs.size();
 	      sout.HLTrigs.push_back(tstruct);
--- a/sbndcode/Decoders/PTB/SBNDPTBDecoder_module.cc
+++ b/sbndcode/Decoders/PTB/SBNDPTBDecoder_module.cc
@@ -207,7 +207,6 @@ void SBNDPTBDecoder::_process_PTB_AUX(const artdaq::Fragment& frag, ptbsv_t &sou
 		tstruct.gate_counter = -1;
 	      }
 
-	      std::cout << tstruct.prev_timestamp << "   ||   " << tstruct.gate_counter << std::endl;
 
 	      ix = sout.HLTrigs.size();
 	      sout.HLTrigs.push_back(tstruct);
--- a/sbndcode/Decoders/PTB/SBNDPTBDecoder_module.cc
+++ b/sbndcode/Decoders/PTB/SBNDPTBDecoder_module.cc
@@ -198,13 +198,13 @@ void SBNDPTBDecoder::_process_PTB_AUX(const artdaq::Fragment& frag, ptbsv_t &sou
 	  if (ctbfrag.Trigger(iword)->IsHLT()) 
 	    {
 	      
-	      if (*(frag.metadata<int>()) == 2){ //If data is taken with 192b PTB words
+	      if ( frag.hasMetadata() == true ){ //If data is past version 2. Make sure to use the right sbndaq-artdaq-core versioning
 		tstruct.prev_timestamp = ctbfrag.PTBWord(iword)->prevTS;
 		tstruct.gate_counter = ctbfrag.Trigger(iword)->gate_counter;
               }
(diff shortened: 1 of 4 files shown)


ups/product_deps
```diff
@@ -346,3 +346,4 @@ table_fragment_end
 ### End:
 
 # vi:set ts=8 noexpandtab:
+
```
bear-is-asleep, line 349:
```suggestion
```

sbndcode/Decoders/PTB/sbndptb.h
```diff
@@ -18,10 +18,10 @@ namespace raw {
   namespace ptb {
 
     struct Trigger {
-      ULong64_t prev_timestamp;
+      long prev_timestamp;
       ULong64_t timestamp;
       ULong64_t trigger_word;
-      uint32_t gate_counter;
+      int gate_counter;
```
terezakroupa, line 24: Ah, I see what my comment for invalid values being set to a negative number messed up... I'd say usage of ULong64_t is better then long for the timestamp, but prevents the unset value to be -1.

I am unsure what's the best solution and happy to defer. Max, can you clarify: for the gate counter, when every gate is accepted, is the gate counter 0 or 1? If it is 1, the 0 is out of range for both the variables and I would just revert this part to the previous version. If 0 is a physical value, then maybe revert the timestamp and keep the gate counter this way?

terezakroupa (commented):
Thank you @maxdubnowski for getting this in. This indeed addresses all my nit-picky comments in the previous PR. Two questions below.

sbndcode/Decoders/PTB/SBNDPTBDecoder_module.cc
```diff
@@ -203,10 +203,12 @@ void SBNDPTBDecoder::_process_PTB_AUX(const artdaq::Fragment& frag, ptbsv_t &sou
 		tstruct.gate_counter = ctbfrag.Trigger(iword)->gate_counter;
               }
               else if (*(frag.metadata<int>()) != 2){ //If data is taken with 128b PTB words
-		tstruct.prev_timestamp = 0;
-		tstruct.gate_counter = 0;
+		tstruct.prev_timestamp = -1;
+		tstruct.gate_counter = -1;
 	      }
 
+	      std::cout << tstruct.prev_timestamp << "   ||   " << tstruct.gate_counter << std::endl;
```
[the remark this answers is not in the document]
bear-is-asleep:
```suggestion
```
I'd vote to remove it

sbndcode/Decoders/PTB/sbndptb.h
lyates17: I think the gate counter should get set to 1 in that case, but it would be great if @maxdubnowski could confirm as well.

maxdubnowski (commented):
Minor adjustments to PTB decoder and analyzer

bear-is-asleep:
@lyates17 and @terezakroupa can we get an approval, or @maxdubnowski can we resolve the issues?

sbndcode/PTBAna/PTBAnalysis_module.cc
```diff
@@ -576,6 +588,7 @@ void sbnd::ptb::PTBAnalysis::AnalysePTBs(std::vector<art::Ptr<raw::ptb::sbndptb>
   _ptb_llt_trigger.resize(nLLTs);
   _ptb_llt_timestamp.resize(nLLTs);
   _ptb_llt_trunmask.resize(nLLTs);
+  _ptb_llt_unmask_timestamp.resize(nHLTs);
```
lyates17, line 591: Should the argument of the `resize()` here be `nLLTs` instead of `nHLTs`? I don't think this actually matters at the moment since these variables are the same but would be good to keep it straight in general.

lyates17:
I would also like to request that this code be tested over raw files from each of the three eras where the PTB data structure was different. The desired behavior is for the code to not crash and to produce output that is as correct as possible considering the information that's available in the DAQ data stream.

- [ ] Example of version 0: run 14500 (from July 2024 beam data-taking)
- [ ] Example of version 1: run 18000 (from December 2024 beam data-taking with DAQ version v1_10_03)
- [ ] Example of version 2: run 18250 (from January 2024 with DAQ version v1_10_04 + Max's latest updates)

I can help track down files from each of these runs if that's useful.

terezakroupa (commented):
Agree with Lauren's comment as something to fix quickly. 

Otherwise it seem all previous comments were resolved and additional changes based on @fjnicolas suggestions were implemented. Looking over the logic it seems good, but worth a check on data (unless it was checked already, Max?).

sbndcode/PTBAna/PTBAnalysis_module.cc
maxdubnowski: Agreed, must have been a copy/paste error

lyates17:
All comments (including my request to run over files from the three different epochs of SBND data with respect to the PTB data format) have now been addressed. So I think this PR is ready to be approved. @bear-is-asleep, can you let us know if there's anything else you need us to before this gets merged.

terezakroupa (approved):

bear-is-asleep:
trigger build LArSoft/lar*@LARSOFT_SUITE_v10_04_04 SBNSoftware/sbncode@v10_04_04

bear-is-asleep:
trigger build LArSoft/lar*@LARSOFT_SUITE_v10_04_03 SBNSoftware/sbnanaobj@v09_23_03 SBNSoftware/sbnobj@v10_00_04 SBNSoftware/sbncode@v10_04_03

PR #766, opened 2025-07-21
Add wc_device jsonnet parameter and gpu_fcls

nathanielerowe:
Just wanted to add this to Mun's feature branch to avoid merge conflicts from trying to merge both into develop at the same time. This enables variable device selection in the DFT and ts model loading. Set using the same methods as was done in icaruscode. 

Have confirmed that standard_detsim_sbnd.fcl still works and the GPU activity is non-zero when running the appropriate fcls. If there is any other kind of validation requested, let me know!

Screenshot from 2025-07-21 15-50-19


wjdanswjddl (approved):
looks good to me -- thanks for putting this together

PR #830, opened 2025-09-22
Feature/trigger work nikki pr

npallat:
## Description 
This includes updates to trigger code to include trigger emulation. Specifically this includes a new LArService that is called in both data and Monte Carlo analysis modules. The LArService include two functions originally from the data analyzer module (ConstructMonPulse and ConstructBinary response) and puts them in this general step that can be applied to both data and Monte Carlo. The functions have also been modified to work with data and Monte Carlo separately where needed. The data analyzer module now simply calls this LArService and the output matches the previous version. The Monte Carlo analyzer module calls this LArService too and now has an interface (in the analyzer module) that slices up the waveforms based on the triggers and consistent with the shorter 10us format of the data waveforms. These new triggered waveforms, the old triggered waveforms, a boolean of if the event passed/failed the trigger, and information to reconstruct the result of ConstructMonPulse (for the purposes of debugging during validation).

## Checklist
- [x] Does this PR affect CAF data format? If so, please assign a CAF maintainer ([PetrilloAtWork](https://github.com/PetrilloAtWork) or [JosiePaton](https://github.com/JosiePaton)) as additional reviewer. _[These changes will affect CAF data format but I am creating a separate PR for that as instructed by Linyan]_
- [x] Does this affect the standard workflow?  _[No]_
- [x] Is this PR a patch for the ongoing production? If so, separate PR must also be made for production/v10_06_00 branch! _[These changes are for the validation production, not the ongoing production/v10_06_00 ]_

Here are a couple relevant links to presentations on this work:
- Update at Trigger Workshop: [DocDB-43231](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=43231)
- Update at Simulation meeting: [DocDB-42969](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=42969)


sbndcode/OpDetSim/TriggerEmulationService.cc
```diff
@@ -0,0 +1,191 @@
...
+ *
+ */
+
+#include "TriggerEmulationService.h"
+
+//namespace lar {
+namespace calib {
+
+  TriggerEmulationService::TriggerEmulationService(fhicl::ParameterSet const& pset, art::ActivityRegistry& amp)
+    : fMonWidth(pset.get<int>("MonWidth", 3)),
```
Jjm321814, line 17: We should adjust these default values to match the data ones, so fMonWidth=12, fTotalCAENBoards=8, PMTPerBoard=15

sbndcode/OpDetSim/TriggerEmulationService.cc
```diff
@@ -0,0 +1,191 @@
...
+              int CAENChannel = 0;
+              // Loop over each PMT in a board
+              while (CAENChannel < PMTPerBoard) {
+
+                  int ChannelStep = 1;
+                  std::vector<bool> BinaryMonContrib(ReadoutSize);
+
+                  if (CAENChannel != 14) {
+                      int WaveIndex_Pair1 = CAENChannel + FlashCounter*PMTPerBoard + CurrentBoard*PMTPerBoard*NumFlash;
+                      int WaveIndex_Pair2 = CAENChannel + 1 + FlashCounter*PMTPerBoard + CurrentBoard*PMTPerBoard*NumFlash;
```
Jjm321814, line 134: Could we make this 
int WaveIndex_Pair2 = WaveIndex_Pair1 + 1; 
so its more explicit and obviously only 1 channel higher

sbndcode/OpDetSim/TriggerEmulationService.cc
```diff
@@ -0,0 +1,191 @@
...
+      int FlashCounter,
+      int *numPairsOverThreshold
+  )
+  {
+      // Loop over the entries in our waveform vector
+      // We care about getting the pairing correct
+
+      std::fill(MonPulse->begin(), MonPulse->end(), 0);    
+
+      if (fMC) { // monte carlo
```
Jjm321814, line 41: So this MC code block is perfectly suitable for running over the initial slicing of MC waveforms (when we get 1 3ms long waveform per PMT). 
After slicing, if we wanted to construct the MonPulse again for saving/plotting, we are going to have issues. It will only run over the first flash. -- Maybe that's okay given that the MonPulse is saved by the OpDetDigitizerSBND_module. Its weird that we would need to reconstruct the MonPulse manually for data and have it saved for MC though. 

I'd say we have 2 options that make sense to me:
1. (Harder option) -- take this block and put it in a special function called "MC_Slicer" or similar, then adjust the code in ConstructMonPulse to assume a more data-like structure to MC after slicing. That way we can always reconstruct the MonPulse as needed. 
2. (Easier option) -- Save the MonPulse in both MC and data early reco steps. For MC this is done, and for data this would probably make sense to add to the PMTDecoder

sbndcode/OpDetSim/TriggerEmulationService.h
```diff
@@ -0,0 +1,63 @@
...
+#include "art/Framework/Services/Registry/ActivityRegistry.h"
+#include "art/Framework/Principal/Handle.h"
+//#include "canvas/Persistency/Common/Handle.h"
+#include "lardataobj/RawData/OpDetWaveform.h"
+#include "larcore/CoreUtils/ServiceUtil.h"
+
+// C/C++ standard libraries
+#include <string>
+#include <vector>
+
```
Jjm321814, line 27: #include <algorithm> to use the max_element suggestion I had in the .cc file

sbndcode/OpDetSim/opdetdigitizer_sbnd.fcl
```diff
@@ -11,6 +11,13 @@ sbnd_opdetdigitizer:
   InputModule:        "PDFastSim"
   WaveformSize:      	    2000.0    #ns (dummy value, resized according to readout of each PD)
   UseSimPhotonsLite:            true  # false for SimPhotons
+  ApplyTriggers:             true  #optional
+  ticksPerSlice:             5000 # corresponds to 10us
+  PercentTicksBeforeCross:             0.2
+  MonThreshold:             15
+  MonPulseThresh:             15
```
Jjm321814, line 18: One of these is an ADC channel value threshold to add 1 to the MON pulse. We should call that a MonPulseThreshold. 
The other one is a pair count threshold for event/flash triggers. We should call that a PairMultiplicityThreshold. 
As named now I can't 100% say which is which just from the name

sbndcode/OpDetSim/opDetDigitizerSBND_module.cc
```diff
@@ -294,42 +363,161 @@ namespace opdet {
...
+          // put trigger pair result in the event
+          auto pairFlag = std::make_unique<int>(numPairsOverThreshold);
+          e.put(std::move(pairFlag), "pairsOverThreshold");
+
+
+          // put sliced waveforms in the event
+          std::unique_ptr< std::vector< raw::OpDetWaveform > > SlicedWaveformsPtr(std::make_unique< std::vector< raw::OpDetWaveform > > ());
+          for (std::vector<raw::OpDetWaveform> &waveforms : fSlicedWaveformsAll) {
+            // move sliced waveforms into the SlicedWaveformsPtr
+            SlicedWaveformsPtr->reserve(SlicedWaveformsPtr->size() + waveforms.size());
```
Jjm321814, line 498: It might be faster to determine the size earlier then do one call to reserve, but only fix this if we see a large increase in DetSim run time, as this is more readable

sbndcode/OpDetSim/opDetDigitizerSBND_module.cc
```diff
@@ -345,6 +533,171 @@ namespace opdet {
...
+                              // if overlaps, extend interval
+                              interestIntervals.back() = {interestIntervals.back().first, crossingPoints[j]+ticksAfterCross};
+                          // if does not overlap, use typical interval length
+                          } else interestIntervals.emplace_back(crossingPoints[j]-ticksBeforeCross, crossingPoints[j]+ticksAfterCross);
+                      // if first, use typical interval length
+                      } else interestIntervals.emplace_back(crossingPoints[j]-ticksBeforeCross, crossingPoints[j]+ticksAfterCross);
+                  }
+              }
+
+              std::vector<std::vector<int>> fSlicedMonPulses;
```
Jjm321814, line 589: We usually use the f<varName> to denote fcl parameters. Maybe we should rename this to something else? But I will defer to Linyan or Nate's opinion on this

Jjm321814:
There is one bit of strange behavior I want to look into in data, with the large MonPulse multiplicity at the tail end of the waveforms. 

This might be being caused by the default values being used, but I don't understand it yet.

sbndcode/OpDetSim/opDetDigitizerSBND_module.cc
npallat: Should we keep this as is for now?

nathanielerowe:
running CI to catch anything not mentioned here that might be broken

nathanielerowe:
trigger build ci_ref=v10_10_02 LArSoft/lar*@LARSOFT_SUITE_v10_10_03 SBNSoftware/sbndaq_artdaq_core@v1_10_06 SBNSoftware/sbnd_data@v01_36_00

sbndcode/OpDetSim/opDetDigitizerSBND_module.cc
Jjm321814: This block is fine as is for now

sbndcode/OpDetSim/TriggerEmulationService.cc
npallat: Note: I am working on change 2 (Easier option) -- this does not impact MC just the data side of things

Jjm321814:
I am happy as soon as we verify the decoders can still run over raw data with the new changes

nathanielerowe:
trigger build ci_ref=v10_10_02 LArSoft/lar*@LARSOFT_SUITE_v10_10_03 SBNSoftware/sbndaq_artdaq_core@v1_10_06 SBNSoftware/sbnd_data@v01_36_00

PR #848, opened 2025-10-15
changes for special blip processing

Jjm321814:
fcl changes to facilitate the special blip hit finding chain. 
The main idea is:
1. Do both traditional and DNN ROI finding (with adjusted thresholds for blips in traditional)
2. Run hit finding over both sets of ROI -- this makes a regular gaushit object (tied to DNN) and a new specialblipgaushit object (tied to regular ROI)
3. Drop the traditional ROI recob::wire prior to reco1 saving
4. Do two sets of pandora track construction for each set of hit finding
5. Point the blip reco at the new extra hit object and extra pandora track object
In blip land we have some arguments about what to do with the middle plane ROI threshold value, so we will do one more update of the space point fcl in WireCell soon.

### DIFF
--- a/sbndcode/CRT/CRTAna/CRTAnalysis_module.cc
+++ b/sbndcode/CRT/CRTAna/CRTAnalysis_module.cc
@@ -1748,9 +1748,9 @@ void sbnd::crt::CRTAnalysis::AnalyseTPCMatching(const art::Event &e, const art::
           _tpc_tr_ts1[nActualTracks]     = crttrack->Ts1();
           _tpc_tr_score[nActualTracks]   = trackMatch.TriggerConfidence();
 
-          const std::set<CRTTagger> taggers = crttrack->Taggers();
-          const geo::Point_t start          = crttrack->Start();
-          const geo::Point_t end            = crttrack->End();
+          const std::vector<CRTTagger> taggers = crttrack->Taggers();
+          const geo::Point_t start             = crttrack->Start();
+          const geo::Point_t end               = crttrack->End();
 
           _tpc_tr_taggers[nActualTracks] = std::vector<int>();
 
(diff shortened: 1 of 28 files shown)


sbndcode/JobConfigurations/standard/reco/config/drops_reco1.fcl
```diff
@@ -12,9 +12,10 @@ sbnd_reco1_ml_drops: [
 sbnd_reco1_drops: [ 
     @sequence::detsim_drops
   , "drop raw::RawDigits_*_*_*"
-  , "drop raw::OpDetWaveforms_*_*_*" 
+  #, "drop raw::OpDetWaveforms_*_*_*"
```
linyan-w: Why comment out dropping OpDetWaveforms?

sbndcode/JobConfigurations/standard/reco/config/drops_reco1.fcl
```diff
@@ -34,6 +35,13 @@ sbnd_reco1_drops_selective: [  #Needed to maintain earlier workflows where some
   , "drop *_cluster3d_*_*" #drop all mlreco output
 ]
 
+sbnd_reco1_drops_data: [ #data specific drops to conserve space where needed
```
linyan-w: Why do we need a data special drop?

nathanielerowe:
@linyan-w can you take a look at this?

linyan-w:
> @linyan-w can you take a look at this?

I did. A few comments pending Jacob to address. Also @Jjm321814 lemme know when you think it's ready and I will test the workflow :D

nathanielerowe:
trigger build LArSoft/lar*@LARSOFT_SUITE_v10_11_01

nathanielerowe:
trigger build LArSoft/lar*@LARSOFT_SUITE_v10_11_01

PR #885, opened 2025-11-24
Trigger Emulation Bug Fix and smaller updates

npallat:
## Description 
Trigger Emulation Bug Fix includes: 
- Bug fix: previously the MonPulse length was taken from the FIRST waveforms in the vector of OpDetWaveforms, but this length corresponded to an XArapuca waveform, which is not the same size. This is fixed to now take the length from the first PMT waveform. (this is the major fix!)
- Added check: added warnings to make sure we are only considering and slicing PMT waveforms
- Added a vector of the number of pairs above threshold to the event (could be added to the cafs later)
- Added a vector of the number of pairs above threshold to the PMT decoder output too
- Updated Baseline fcl parameter to Run 2 Baseline,

$${\color{red}!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!}$$

$${\color{red}\bf{\textrm{IMPORTANT UPDATE June 22nd 2025:}}}$$  If you are making a PR which is intended as a patch for the CURRENT production (which started in Spring 2025), you must make two PRs: one for develop and one for the production/v10_06_00 branch.

$${\color{red}!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!}$$

## Checklist
- [x] Does this PR affect CAF data format? If so, please assign a CAF maintainer ([PetrilloAtWork](https://github.com/PetrilloAtWork) or [JosiePaton](https://github.com/JosiePaton)) as additional reviewer.
- [x] Does this affect the standard workflow? 
- [x] Is this PR a patch for the ongoing production? If so, separate PR must also be made for production/v10_06_00 branch! 
This affects the ongoing validation release. 

It does not require merging with another PR. However, here is a link to the [original PR](https://github.com/SBNSoftware/sbndcode/pull/830/files#diff-6d6eaa11a42da81d523815e69fc004f0f8a2fb93cab5a859c11a38cae67738d2) including this work, for reference.

Here is a relevant [presentation](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=44542) about this to the Trigger Working Group meeting.

### DIFF
--- a/sbndcode/Decoders/PMT/run_pmtdecoder.fcl
+++ b/sbndcode/Decoders/PMT/run_pmtdecoder.fcl
@@ -11,7 +11,7 @@ services: {
     MonWidth: 12
     TotalCAENBoards: 8
     PMTPerBoard: 15
-    Baseline: 14250
+    Baseline: 14257 #Run 1: 14250, Run 2: 14257
     MC: false
   }
 }
--- a/sbndcode/Decoders/PMT/run_pmtdecoder.fcl
+++ b/sbndcode/Decoders/PMT/run_pmtdecoder.fcl
@@ -11,7 +11,7 @@ services: {
     MonWidth: 12
     TotalCAENBoards: 8
     PMTPerBoard: 15
-    Baseline: 14257 #Run 1: 14250, Run 2: 14257
+    Baseline: 14250 #Run 1: 14250, Run 2: 14257
     MC: false
   }
 }
--- a/sbndcode/OpDetSim/TriggerEmulationService.cc
+++ b/sbndcode/OpDetSim/TriggerEmulationService.cc
@@ -17,7 +17,7 @@ namespace calib {
     : fMonWidth(pset.get<int>("MonWidth", 12)),
       fTotalCAENBoards(pset.get<int>("TotalCAENBoards", 8)),
       PMTPerBoard(pset.get<int>("PMTPerBoard", 15)),
-      Baseline(pset.get<int>("Baseline", 14257)),
+      Baseline(pset.get<int>("Baseline")),
       fMC(pset.get<bool>("MC", true))
   {}
 
@@ -69,10 +69,10 @@ namespace calib {
               if (PMT_Channels.empty()) std::cout<<"Warning: Please provide PMT channels list to check if channels are PMT channels."<<std::endl;
               else {
                   if (!(std::find(PMT_Channels.begin(), PMT_Channels.end(), ch1) != PMT_Channels.end())) {
-                      std::cout<<"Paired channel "<<ch1<<" is not PMT channel. Check Pairs list. Skipping..."<<std::endl;
+                      throw cet::exception("TriggerEmulationService") << "Paired channel " << ch1 << " is not PMT channel. Check Pairs list. Skipping..." << std::endl;
                       continue; 
                   } if (!(std::find(PMT_Channels.begin(), PMT_Channels.end(), ch2) != PMT_Channels.end())) { 
-                      std::cout<<"Paired channel "<<ch2<<" is not PMT channel. Check Pairs list. Skipping..."<<std::endl; 
+                      throw cet::exception("TriggerEmulationService") << "Paired channel " << ch2 << " is not PMT channel. Check Pairs list. Skipping..." << std::endl;
                       continue; 
                   }
               }
@@ -104,7 +104,7 @@ namespace calib {
               if (PMT_Channels.empty()) std::cout<<"Warning: Please provide PMT channels list to check if channels are PMT channels."<<std::endl;
               else {
                   if (!(std::find(PMT_Channels.begin(), PMT_Channels.end(), ch) != PMT_Channels.end())) {
-                      std::cout<<"Unpaired channel "<<ch<<" is not PMT channel. Check list. Skipping..."<<std::endl;
+                      throw cet::exception("TriggerEmulationService") << "Unpaired channel " << ch << " is not PMT channel. Check list. Skipping..." << std::endl;
                       continue; 
                   }
               }
--- a/sbndcode/OpDetSim/TriggerEmulationService.cc
+++ b/sbndcode/OpDetSim/TriggerEmulationService.cc
@@ -17,7 +17,7 @@ namespace calib {
     : fMonWidth(pset.get<int>("MonWidth", 12)),
       fTotalCAENBoards(pset.get<int>("TotalCAENBoards", 8)),
       PMTPerBoard(pset.get<int>("PMTPerBoard", 15)),
-      Baseline(pset.get<int>("Baseline", 14250)),
+      Baseline(pset.get<int>("Baseline", 14257)),
       fMC(pset.get<bool>("MC", true))
   {}
 
@@ -29,9 +29,11 @@ namespace calib {
       int MonThreshold,
       std::vector<int> *MonPulse,
       int FlashCounter,
-      int *numPairsOverThreshold
+      int *numPairsOverThreshold,
+      std::vector<int> PMT_Channels
   )
   {
+
       // Loop over the entries in our waveform vector
       // We care about getting the pairing correct
 
@@ -44,8 +46,9 @@ namespace calib {
           }
 
           std::map<int, const raw::OpDetWaveform*> channel_to_waveform;
-          for (const auto& wvf : fWaveforms)
+          for (const auto& wvf : fWaveforms) {
               channel_to_waveform[wvf.ChannelNumber()] = &wvf;
+          }
 
           std::vector<int> Pair2 = { 6,   8,  10,  12,  14,  16,  36,  38,  40,  60,  62,  66,  68, 70,  84,  86,  88,  90,  92,  94, 114, 116, 118, 138, 140, 144, 146, 148, 162, 164, 168, 170, 172, 192, 194, 196, 216, 218, 220, 222, 224, 226, 240, 242, 246, 248, 250, 270, 272, 274, 294, 296, 298, 300, 302, 304};
           std::vector<int> Pair1 = { 7,   9,  11,  13,  15,  17,  37,  39,  41,  61,  63,  67,  69, 71,  85,  87,  89,  91,  93,  95, 115, 117, 119, 139, 141, 145, 147, 149, 163, 165, 169, 171, 173, 193, 195, 197, 217, 219, 221, 223, 225, 227, 241, 243, 247, 249, 251, 271, 273, 275, 295, 297, 299, 301, 303, 305};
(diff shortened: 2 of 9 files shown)


sbndcode/Decoders/PMT/run_pmtdecoder.fcl
fjnicolas, line 1: We should leave the baseline at 14250 ADCs since we will be processing run 1 data, but this is definitely something to take into account when we start processing run 2. We could open an issue to remind ourselves in the future.

sbndcode/OpDetSim/TriggerEmulationService.cc
```diff
@@ -17,7 +17,7 @@ namespace calib {
     : fMonWidth(pset.get<int>("MonWidth", 12)),
       fTotalCAENBoards(pset.get<int>("TotalCAENBoards", 8)),
       PMTPerBoard(pset.get<int>("PMTPerBoard", 15)),
-      Baseline(pset.get<int>("Baseline", 14250)),
+      Baseline(pset.get<int>("Baseline", 14257)),
```
fjnicolas: Same here, we want to keep run1 baseline value by now. In general, it is good practice not giving default values in the constructor so we make sure everything is set up from the fhicl.
```suggestion
      Baseline(pset.get<int>("Baseline")),
```

sbndcode/OpDetSim/run_BeamRatesCalib.fcl
```diff
@@ -17,7 +17,7 @@ services:
     MonWidth: 12
     TotalCAENBoards: 8
     PMTPerBoard: 15
-    Baseline: 14250
+    Baseline: 14257 #Run 1: 14250, Run 2: 14257
```
fjnicolas: Leave run 1 baseline
```suggestion
    Baseline: 14250 #Run 1: 14250, Run 2: 14257
```

sbndcode/OpDetSim/triggeremulationservice_sbnd.fcl
```diff
@@ -6,7 +6,7 @@ sbnd_triggeremulation_service: {
   MonWidth: 12
   TotalCAENBoards: 8
   PMTPerBoard: 15
-  Baseline: 14250
+  Baseline: 14257 #Run 1: 14250, Run 2: 14257
```
fjnicolas: Leave run 1 baseline
```suggestion
  Baseline: 14250 #Run 1: 14250, Run 2: 14257
```

sbndcode/OpDetSim/opDetDigitizerSBND_module.cc
```diff
@@ -373,65 +374,67 @@ namespace opdet {
           art::ServiceHandle<art::TFileService> tfs;
           art::ServiceHandle<calib::TriggerEmulationService> fTriggerService;
 
-          int PMTPerBoard = fTriggerService->getPMTPerBoard();
-          int fTotalCAENBoards = fTriggerService->getTotalCAENBoards();
+          std::vector<int> numPairsOverThreshold;
 
-          int TotalFlash = fWaveforms.size() / (fTotalCAENBoards * PMTPerBoard);
+          // get PMT channels
+          std::vector<int> PMT_Channels;
```
fjnicolas: Can you move this part to the class constructor? As it is right now we are repeating the same loop for every event.

sbndcode/OpDetSim/opDetDigitizerSBND_module.cc
```diff
@@ -373,65 +374,67 @@ namespace opdet {
...
-              if (SaveNewPlots) {
-                  // Save histograms
-                  // Sliced waveforms
-                  for (size_t j; j < SlicedWaveformsAll.size(); ++j) { 
-                      std::stringstream plotname2; 
-                      plotname2 << "Sliced_waveforms_" << e.id().event() << "_Mon_" << MonThreshold << "_" << FlashCounter << "_slice" << j;
-                      PlotWaveforms(SlicedWaveformsAll[j], plotname2.str());
-                  }
-                  // Long MonPulse
+          if (SaveNewPlots) {
```
fjnicolas: I just saw it is using the TFileService to make histograms with the MON pulses. Why is that the case? I'd rather having this part in a separated analyzer module instead of doing inside of the main producer module that is used in production (it's probably adding unnecessary runtime/file size to the standard SBND workflow).

sbndcode/OpDetSim/opDetDigitizerSBND_module.cc
```diff
@@ -373,65 +374,67 @@ namespace opdet {
...
-          int TotalFlash = fWaveforms.size() / (fTotalCAENBoards * PMTPerBoard);
+          // get PMT channels
+          std::vector<int> PMT_Channels; 
+          for (const auto& wvf : fWaveforms) {
+            auto ch = wvf.ChannelNumber();
+            if (map.isPDType(ch, "pmt_uncoated") || map.isPDType(ch, "pmt_coated")) PMT_Channels.push_back(ch);
+          }
 
-          int numPairsOverThreshold = 0;
+          if (PMT_Channels.empty()) std::cout<<"Error: PMT has *NO* PMT channels in fWaveforms"<<std::endl;
```
fjnicolas: This one should be an exception instead.
`throw cet::exception("opDetDigitizerSBND") << "PMT has *NO* PMT channels in fWaveforms" << std::endl;`

sbndcode/OpDetSim/TriggerEmulationService.cc
```diff
@@ -44,22 +46,37 @@ namespace calib {
...
 
           for (size_t i = 0; i < Pair1.size(); ++i) {
               int ch1 = Pair1[i];
               int ch2 = Pair2[i];
 
+              // check that ch1 and ch2 show up in PMT_Channels
+              if (PMT_Channels.empty()) std::cout<<"Warning: Please provide PMT channels list to check if channels are PMT channels."<<std::endl;
+              else {
+                  if (!(std::find(PMT_Channels.begin(), PMT_Channels.end(), ch1) != PMT_Channels.end())) {
+                      std::cout<<"Paired channel "<<ch1<<" is not PMT channel. Check Pairs list. Skipping..."<<std::endl;
```
fjnicolas: This one (and the ones following) should probably throw an exception if the PMT channels provided do not match the pairing scheme.

sbndcode/OpDetSim/TriggerEmulationService.cc
```diff
@@ -44,22 +46,37 @@ namespace calib {
...
 
           std::vector<int> Pair2 = { 6,   8,  10,  12,  14,  16,  36,  38,  40,  60,  62,  66,  68, 70,  84,  86,  88,  90,  92,  94, 114, 116, 118, 138, 140, 144, 146, 148, 162, 164, 168, 170, 172, 192, 194, 196, 216, 218, 220, 222, 224, 226, 240, 242, 246, 248, 250, 270, 272, 274, 294, 296, 298, 300, 302, 304};
           std::vector<int> Pair1 = { 7,   9,  11,  13,  15,  17,  37,  39,  41,  61,  63,  67,  69, 71,  85,  87,  89,  91,  93,  95, 115, 117, 119, 139, 141, 145, 147, 149, 163, 165, 169, 171, 173, 193, 195, 197, 217, 219, 221, 223, 225, 227, 241, 243, 247, 249, 251, 271, 273, 275, 295, 297, 299, 301, 303, 305};
           std::vector<int> Unpaired = {65,  64, 143, 142, 167, 166, 245, 244};
           std::set<int> used_channels;
 
           // resize
-          int ReadoutSize = fWaveforms[0].size();
+          int ReadoutSize;
+          if (PMT_Channels.empty()) { std::cout<<"Warning: Please provide PMT channels list."<<std::endl; ReadoutSize = fWaveforms[Pair1[0]].size(); }
```
fjnicolas, line 60: Is the cout here really a warning? It is expected that the PMT_Channels vector is empty when you call it fom the PMTDecoder module.

fjnicolas (changes requested):
Thanks @npallat for putting this in!

My request before this gets in is reverting back to use run 1 baseline, since currently all our processing is based on run 1.

I left some other suggestions that I think should get address before a full scale production, but given the time constraint for the validation II samples I'd rather getting this in by now. Any thoughts @nathanielerowe @linyan-w?

sbndcode/OpDetSim/run_BeamRatesCalib.fcl
asanchezcastillo: Hi! Just chiming in, the baseline value should be the same as the one used in (https://github.com/SBNSoftware/sbndcode/blob/develop/sbndcode/OpDetSim/digi_pmt_sbnd.fcl#L27), so we should probably use fhicl inheritance to make sure we are always taking the same baseline. For PMT detsim we are currently using 14750 (not sure where that's coming from honestly) but we should probably change it to a data-based baseline.

sbndcode/OpDetSim/TriggerEmulationService.cc
npallat: The PMTDecoder has to assign the ReadoutSize differently because it is using data not MC:
```
          int NumFlash = fWaveforms.size() / (fTotalCAENBoards * PMTPerBoard);
          int FirstReadoutIndex = 0 + FlashCounter*PMTPerBoard + 0*PMTPerBoard*NumFlash;
          int ReadoutSize = fWaveforms[FirstReadoutIndex].size();
```
I could change this to just assign the ReadoutSize as `ReadoutSize = fWaveforms[Pair1[0]].size()` though since a warning will appear later when we loop through the channels if PMT_Channels is empty.

sbndcode/OpDetSim/run_BeamRatesCalib.fcl
npallat: @fjnicolas It seems the PMT detsim is using 14750 and digi_pmt_sbnd.fcl is using 14745, while Run 1 data is using 14250. Should I change the Baseline to 14745?

I definitely agree with @asanchezcastillo we should change this to use fhicl inheritance before the full production.

sbndcode/Decoders/PMT/run_pmtdecoder.fcl
npallat: I have addressed all of the comments except moving the "get PMT channels" loop to the constructor. I am fine with this being opened as an issue and making this change before full scale production.
I waiting to push these changes until I get a confirmation that we would like to use 14250 as the Baseline though, given Alejandro's comments.

fjnicolas:
thanks for the catch @asanchezcastillo! my suggestion then is:

- changing the baseline in `digi_pmt_sbnd.fcl` to use 14250 ADC (as in run 1 data)
- TriggerEmulation inherits the baseline value from the PMT fhicl

Otherwise we'd need to have two configurations for the TriggerEmulation service, one for data, one for MC.

fjnicolas (approved):
Thank you so much Nikki! Ready for CI!

nathanielerowe:
trigger build LArSoft/lar*@LARSOFT_SUITE_v10_14_00 SBNSoftware/sbndaq-artdaq-core@v1_10_06 SBNSoftware/sbn*@SBN_SUITE_v10_14_00

nathanielerowe:
trigger build LArSoft/lar*@LARSOFT_SUITE_v10_14_00 SBNSoftware/sbndaq-artdaq-core@v1_10_06 SBNSoftware/sbn*@SBN_SUITE_v10_14_00

nathanielerowe:
Approved

PR #901, opened 2026-01-20
Bugfix/blips on empty events

Jjm321814:
A few bugs existed in old versions of blip reco

1. Running blipreco on mostly empty event displays (somewhat common in simple MC productions of single neutrino events) would lead to empty hit collections being accessed and a subsequent memory issue/crash (Not specifically a segfault though). 
2. A few blip-related hitcluster variables were not copying quite correctly. Effected variables are blip.clusters[i].blipID, and blip.clusters[i].isMatched, which both had default values. The independently saved clusters objects were correctly updated. 
3. The MC truth PDG values for blip-energy deposits have puzzling labels and need their evaluation looked into. 

The first bug was related to a hardcoded hitHandleGH label, which looked at gaushit rather than specialBlipGausHit. On ~emtpy events the lower threshold (specialBlipGausHit) will have a few entries, but the code uses hitHandleGH for truth matching, and that higher threshold hit collection can be empty. That leads to crashes. 
For now the hardcoded label is just updated to accommodate specialBlipGausHit. 

The second bug is not critical, as access hitclusters this way is done through a blip, so the ID is the same as the blip you are looking at and the clusters included must be matched (into this blip). The hitcluster collection values are updated after the struct assignment (copy constructor), so I updated the blip alg to also update the blip.Cluster[i].blipID, and isMatched. 

The third bug I am still digging into. I was worried that the switch from sim::energy_deposition to sim::channel->IDE was causing issues, but in the new arrangement (with an optional interface to use either) gives consistent results between them. 

I have some notes I can build into slides for the first two but would like to spend a little more time on the third bug.

### DIFF
--- a/sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
+++ b/sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
@@ -357,31 +357,32 @@ namespace blip {
     if (evt.getByLabel(fGeantProducer,pHandle))
       art::fill_ptr_vector(plist, pHandle);
  
-    // -- SimEnergyDeposits (usually dropped in reco
-    //art::Handle<std::vector<sim::EnergyDeposit> > sedHandle;
-    std::vector<sim::IDE > sedlist;
-    //if (evt.getByLabel(fSimDepProducer,sedHandle)){
-    //  art::fill_ptr_vector(sedlist, sedHandle);
-    // }
-    // -- SimChannels (usually dropped in reco)
+    // -- SimEnergyDeposits 
+    art::Handle<std::vector<sim::EnergyDeposit> > sedHandle;
+    std::vector<art::Ptr<sim::SimEnergyDeposit> > sedlist;
+    if (evt.getByLabel(fSimDepProducer,sedHandle)){
+      art::fill_ptr_vector(sedlist, sedHandle);
+     }
+    std::vector<sim::IDE > sIDElist;
+    // -- SimChannels 
     art::Handle<std::vector<sim::SimChannel> > simchanHandle;
     std::vector<art::Ptr<sim::SimChannel> > simchanlist;
     if (evt.getByLabel(fSimChanProducer,simchanHandle))
     { 
       art::fill_ptr_vector(simchanlist, simchanHandle);
-      //Loop over channels to get full sedlist
+      //Loop over channels to get full sIDElist
       for(int chIndex=0; chIndex<int(simchanlist.size()); chIndex++)
 	    {
       std::vector<geo::WireID> wids    = wireReadoutGeom->Get().ChannelToWire( (*(simchanlist[chIndex])).Channel() ); //Not sure why this is a vector, but it should have len 1
-      std::cout << " Got my wire for chIndex " << chIndex << " it has size " << wids.size() << " first entry string is " << wids[0].toString() << std::endl;
       const geo::PlaneID&      planeID = wids[0].planeID();
       if(int(planeID.Plane) != fCaloPlane) continue; //only take calorimetry plane IDE values 
-      std::vector< sim::IDE > TempChIDE = (*simchanlist[chIndex]).TrackIDsAndEnergies(0, 999999999);
+      std::cout << " Got my wire for chIndex " << chIndex << " it has size " << wids.size() << " first entry string is " << wids[0].toString() << std::endl;
+      unsigned int MaxTDCTick = 3401;
+      std::vector< sim::IDE > TempChIDE = (*simchanlist[chIndex]).TrackIDsAndEnergies(0, MaxTDCTick);
       std::cout << " this channel has " << TempChIDE.size() << " IDEs" << std::endl;
       for(int ideIndex=0; ideIndex<int(TempChIDE.size()); ideIndex++)
 	        {
-	        //art::fill_ptr_vector(sedlist, simchanHandle.TrackIDsAndEnergies(0, 99999999));
-	        sedlist.push_back( TempChIDE[ideIndex] ); //may need to add a &
+	        sIDElist.push_back( TempChIDE[ideIndex] ); //may need to add a &
 	       }
 	    }
     }
@@ -535,7 +536,15 @@ namespace blip {
     if( plist.size() ) {
       pinfo.resize(plist.size());
       for(size_t i = 0; i<plist.size(); i++){
-        BlipUtils::FillParticleInfo( *plist[i], pinfo[i], sedlist, fCaloPlane);
+        //use sim::EnergyDeposits by default. This is heavy and may be dropped
+        if(evt.getByLabel(fSimDepProducer,sedHandle))
+        {
+          BlipUtils::FillParticleInfo( *plist[i], pinfo[i], sedlist, fCaloPlane);
+        }
+        else //use sim::Channel -> IDE otherwise. This is usually kept but results in strange bugs.
+        {
+          BlipUtils::FillParticleInfo( *plist[i], pinfo[i], sIDElist, fCaloPlane);
+        }
         if( map_g4trkid_charge[pinfo[i].trackId] ) pinfo[i].numElectrons = (int)map_g4trkid_charge[pinfo[i].trackId];
         pinfo[i].index = i;
       }
--- a/sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
+++ b/sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
@@ -32,7 +32,6 @@ namespace blip {
       
 
     // Loop over cryostats
-    std::cout<<"NCryostats: "<<fGeom.Ncryostats()<<"\n";
     for(size_t cstat=0; cstat<fGeom.Ncryostats(); cstat++){
       auto const& cryoid = geo::CryostatID(cstat);
 
@@ -49,8 +48,6 @@ namespace blip {
           kNumChannels += planegeo.Nwires();
             
           float offset = detProp.GetXTicksOffset(pl,tpc,cstat);
-          std::cout<<"CRYOSTAT "<<cstat<<" / TPC "<<tpc<<" / PLANE "<<pl<<":  "<<planegeo.Nwires()<<" wires\n";
-          std::cout<<"  XTicksOffset (from detProp): "<<offset<<"\n";
          
           kXTicksOffsets[cstat][tpc][pl] = 0;
 
@@ -68,10 +65,7 @@ namespace blip {
             const double dir((tpcgeom.DriftSign() == geo::DriftSign::Negative) ? +1.0 : -1.0);
 	    float x_ticks_coefficient = kDriftVelocity*kTickPeriod;
             float goofy_offset = -xyz.X() / (dir * x_ticks_coefficient);
-	    std::cout<<"  After geometric correction: "<<offset - goofy_offset<<"\n";
-
             kXTicksOffsets[cstat][tpc][pl] = offset - goofy_offset;
-
 	  } else {
           
             // for the case of 2D wirecell workflow, the plane-to-plane
@@ -85,9 +79,7 @@ namespace blip {
           }
           
           // additional ad-hoc corrections supplied by user
-          //kXTicksOffsets[cstat][tpc][pl] += fTimeOffset[pl];
-          std::cout << " offsetting plane " << pl << " by " << fTimeOffset[pl] << " ticks " << std::endl;
-          
+          //kXTicksOffsets[cstat][tpc][pl] += fTimeOffset[pl];          
         }
       }
     }
@@ -361,7 +353,6 @@ namespace blip {
     art::Handle<std::vector<sim::SimEnergyDeposit> > sedHandle;
     std::vector<art::Ptr<sim::SimEnergyDeposit> > sedlist;
     if (evt.getByLabel(fSimDepProducer,sedHandle)){
-      std::cout << " in the sedlist filler " << std::endl;
       art::fill_ptr_vector(sedlist, sedHandle);
      }
     std::vector<sim::IDE > sIDElist;
@@ -385,8 +376,6 @@ namespace blip {
 	       }
 	    }
     }
-    std::cout << "sed list " << sedlist.size() << " IDElist " << sIDElist.size() << std::endl;
-
     // -- hits (from input module, usually track-masked subset of gaushit)
     art::Handle< std::vector<recob::Hit> > hitHandle;
     std::vector<art::Ptr<recob::Hit> > hitlist;
@@ -449,9 +438,7 @@ namespace blip {
     //===============================================================
     std::map< int, int > map_gh;
     // if input collection is already gaushit, this is trivial
-    std::cout << " About to fill in map with hitProducer " << fHitProducer << std::endl;
     if( fHitProducer == "gaushit" ||  fHitProducer == "specialblipgaushit") {
-      std::cout << " in special branch" << std::endl;
       for(auto& h : hitlist ) map_gh[h.key()] = h.key(); 
     // ... but if not, find the matching gaushit. There's no convenient
     // hit ID, so we must loop through and compare channel/time (ugh)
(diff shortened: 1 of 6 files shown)


nathanielerowe (approved):

Jjm321814:
https://docs.google.com/presentation/d/1igXPcS_Qst99layiHuMEZiUxz4EHMxGHutf7Mr-jbT0/edit?slide=id.g3bbdbff95f2_0_1#slide=id.g3bbdbff95f2_0_1 here is a useful overview set of slides for these changes. Sorry for the many commits most are just adding/removing debug. 

For the hardcoded gaushit parameter the change in this PR is sufficient for any standard running someone does/for Fall production. Longterm I think these should not be hardcoded and just use the fcl-provided hit collection, and we assume the user knows what they are doing. 

The existing MC truth back tracking works. I just added a few additional breaks in MC truth blips. I found a way to get labeling I am totally happy with but the performance hit is too large for anything other than a small production aimed specifically at blips. 

I am happy with the current state of the MCTruth matching in this PR now. I promise some future slides looking at how the MC matching is doing in Fall production pre/post this PR in the next ~week

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
```diff
@@ -68,10 +65,7 @@ namespace blip {
             const double dir((tpcgeom.DriftSign() == geo::DriftSign::Negative) ? +1.0 : -1.0);
 	    float x_ticks_coefficient = kDriftVelocity*kTickPeriod;
             float goofy_offset = -xyz.X() / (dir * x_ticks_coefficient);
```
henrylay97, line 67: There is lots of bad mis-alignment in these files, should be fixable with a quick de-tabify offline :)

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
```diff
@@ -85,9 +79,7 @@ namespace blip {
           }
           
           // additional ad-hoc corrections supplied by user
-          //kXTicksOffsets[cstat][tpc][pl] += fTimeOffset[pl];
-          std::cout << " offsetting plane " << pl << " by " << fTimeOffset[pl] << " ticks " << std::endl;
-          
+          //kXTicksOffsets[cstat][tpc][pl] += fTimeOffset[pl];
```
henrylay97: Why is this commented and not just removed?

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
```diff
@@ -357,33 +349,33 @@ namespace blip {
...
-	  if(int(planeID.Plane) != fCaloPlane) continue; //only take calorimetry plane IDE values 
-	  std::vector< sim::IDE > TempChIDE = (*simchanlist[chIndex]).TrackIDsAndEnergies(0, 999999999);
-	  for(int ideIndex=0; ideIndex<int(TempChIDE.size()); ideIndex++)
 	    {
-	      //art::fill_ptr_vector(sedlist, simchanHandle.TrackIDsAndEnergies(0, 99999999));
-	      sedlist.push_back( TempChIDE[ideIndex] ); //may need to add a &
+      std::vector<geo::WireID> wids    = wireReadoutGeom->Get().ChannelToWire( (*(simchanlist[chIndex])).Channel() ); //Not sure why this is a vector, but it should have len 1
+      const geo::PlaneID&      planeID = wids[0].planeID();
+      if(int(planeID.Plane) != fCaloPlane) continue; //only take calorimetry plane IDE values 
+      unsigned int MaxTDCTick = 4294967294; // 1 under 32 bit unsigned int max
```
henrylay97: Why 1 less? And you could use UINT_MAX here for better readability and to keep it consistent for any system.

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
```diff
@@ -393,8 +385,8 @@ namespace blip {
     // -- hits (from gaushit), these are used in truth-matching of hits
     art::Handle< std::vector<recob::Hit> > hitHandleGH;
     std::vector<art::Ptr<recob::Hit> > hitlistGH;
-    if (evt.getByLabel("gaushit",hitHandleGH))
-      art::fill_ptr_vector(hitlistGH, hitHandleGH);
+    if(evt.getByLabel("specialblipgaushit",hitHandleGH)) art::fill_ptr_vector(hitlistGH, hitHandleGH);
+    else if(evt.getByLabel("gaushit",hitHandleGH))       art::fill_ptr_vector(hitlistGH, hitHandleGH);
```
henrylay97: Module labels should never be hard coded. I appreciate this was true before this PR but given you're touching the code at this point it needs fixing.

I also don't understand the use of two different hit collections - only one of them will have been an input to the BlipReco and so only one of them is a valid input for the truth matching. If the code crashes on an empty vector somewhere then the fix should be a catch to stop that happening rather than the providing of a different vector.

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
```diff
@@ -405,7 +397,7 @@ namespace blip {
     // -- associations
     art::FindManyP<recob::Track> fmtrk(hitHandle,evt,fTrkProducer);
     art::FindManyP<recob::Track> fmtrkGH(hitHandleGH,evt,fTrkProducer);
-    art::FindMany<simb::MCParticle,anab::BackTrackerHitMatchingData> fmhh(hitHandleGH,evt,"gaushitTruthMatch");
+    art::FindMany<simb::MCParticle, anab::BackTrackerHitMatchingData> fmhh(hitHandleGH,evt,"blipgaushitTruthMatch");
```
henrylay97: Same comment on hard coding

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
```diff
@@ -442,7 +434,7 @@ namespace blip {
     //===============================================================
     std::map< int, int > map_gh;
     // if input collection is already gaushit, this is trivial
-    if( fHitProducer == "gaushit" ) {
+    if( fHitProducer == "gaushit" ||  fHitProducer == "specialblipgaushit") {
```
henrylay97: Why is this necessary?

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
```diff
@@ -647,7 +646,7 @@ namespace blip {
       
 
       // find associated track
-      if( fHitProducer == "gaushit" && fmtrk.isValid() ) {
+      if( fHitProducer == "specialblipgaushit" && fmtrk.isValid() ) {
```
henrylay97: Again same comment, this feels like a solution that creates a lot of future problems. Why should this block only be run if the HitProducer is a certain value?

sbndcode/BlipRecoSBND/BlipAna_module.cc
```diff
@@ -1582,6 +1581,9 @@ void BlipAna::PrintClusterInfo(const blip::HitClust& hc){
     hc.EdepID,
     hc.isMatched
   );
+  printf("G4 IDs contib \n");
+  for(int g4ID : hc.G4IDs) printf(" %i ", g4ID);
+  printf("\n");
```
henrylay97, line 1586: Is BlipAna run as standard in production?

For example, in the calib ntuples? If so, the print statements need to be hidden behind a debug flag.

henrylay97 (changes requested):
Hi Jacob, thanks for the PR. I have a few comments.

My main concern is around the fix for the empty event problem. Rather than finding the root cause of the problem, the solution seems to be to provide a second vector that you hope won't be empty as well.

I would rather the actual cause of the segfault be found and prevented. This would also remove any need for hard coded module labels.

The rest looks good to me, happy to discuss more offline!

sbndcode/BlipRecoSBND/BlipAna_module.cc
Jjm321814: BlipAna is run in standard production for calib ntuples. 
The function these print statements live in is only called with a debug flag
`if( fDebugMode ) PrintClusterInfo(clust);`

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
Jjm321814: Went through and generally fixed tab alignment. There are a few lines of white space with awkward tabbing still, but all the content looks aligned properly on my end.

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
Jjm321814: Good point. The timeOffset is applied independently later to the hits directly. This can be removed.

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
Jjm321814: Didn't know that existed. Updated it now

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
Jjm321814: I have removed the hitHandleGH, and will remove the other spots that gaushit gets a special default privileging. It will just be the user's responsibility to make sure they have ran all the necessary inputs

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
Jjm321814: Added a new fcl param that defaults to blipgaushitTruthMatch

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
Jjm321814: This was some functionality to allow the user to provide a hitcollection that never had truth matching or PandoraTracking ran on it. That was the whole purpose of the hitHandleGH, and I think that was more useful for uB reprocessing and just got included when the blips were ported over to SBND. I've left in map_gh, but this branching is eliminated and it just assumes the trivial case where our hitcollection was ran through truth matching and tracking.

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
Jjm321814: Changed to just check if fmtrk is a valid handle.

Jjm321814:
Thanks for the suggestions, most are in. Let me check compilation and a simple output in a typical Fall production run

Jjm321814:
Merged in develop and the compiler works and entries in the reco2 files look good.

After getting the normal successful processing table I got this error
*** Error in `lar': corrupted size vs. prev_size: 0x000000002a0308b0 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x80c37)[0x7fb0b2cb2c37]
/lib64/libc.so.6(+0x8120e)[0x7fb0b2cb320e]
/lib64/libc.so.6(+0x39ce9)[0x7fb0b2c6bce9]
/lib64/libc.so.6(+0x39d37)[0x7fb0b2c6bd37]
/lib64/libc.so.6(__libc_start_main+0xfc)[0x7fb0b2c5455c]
lar[0x4012d0]
Followed by a long memory map. I don't think this is my fault.

Jjm321814:
Fixed the above error.
After merging in develop it became v10_14_02_01 and my localproducts setup script was still asking for v10_14_02. Fixing that removed the error on exit.

henrylay97:
Thanks so much for responding so quickly. Looking much better.

I have a question. Will this now not segfault on empty events? Regardless of the label being applied.

I note you say: "_I have removed the hitHandleGH, and will remove the other spots that gaushit gets a special default privileging. It will just be the user's responsibility to make sure they have ran all the necessary inputs_"

Does this mean if the user mislabels the input we get a segfault? If so we need to avoid that and instead have a more graceful failure mode :)

Jjm321814:
I'll check that now. Noting down the possible combinations of wrong label:
1. Hitlabel doesn't exist
2. Track label and hit label are from different sets (regular pandora and special blip hits for example)
3. MC Truth Matching done on a different hit set
4. Track label doesn't exist
5. MC truth label doesn't exist
I'll check each combination and report back how it exits

Jjm321814:
We don't ever segfault. But we do have some cases where a normal output will be produced and the user must be careful that the fcl used to make it was correct. 

For case 1 
```
%MSG-s ArtException:  PostEndJob 29-Jan-2026 09:25:51 CST ModuleEndJob
---- EventProcessorFailure BEGIN
  EventProcessor: an exception occurred during current event processing
  ---- ScheduleExecutionFailure BEGIN
    Path: ProcessingStopped.
    ---- ProductNotFound BEGIN
      Found zero products matching all selection criteria
        C++ type: std::vector<recob::Hit>
        Module label: 'specialblipgaushitIdontexist'
        Product instance name: ''
        Process name: (empty)
      The above exception was thrown while processing module BlipRecoProducer/blipreco run: 1001 subRun: 18 event: 1
    ---- ProductNotFound END
    Exception going through path reco2
  ---- ScheduleExecutionFailure END
---- EventProcessorFailure END
---- FatalRootError BEGIN
  Fatal Root Error: TTree::SetEntries
  Tree branches have different numbers of entries, eg EventAuxiliary has 0 entries while sim::AuxDetSimChannels_genericcrt__G4. has 13 entries.
  ROOT severity: 2000
---- FatalRootError END
```

For case 2 
The code seems to just run normally. The track input is really just to skip a few hits that are not-blip-like. Different indexing in the two track associations may cause some hits that are really blips to be skipped. This would be a tricky bug to catch so its important the fcl are ironed out. 

For case 3:
Runs normally. Inspection of output should give you some clue that things aren't working well. But again this is a be careful with fcl case. 

For case 4:
This one actually just runs normally and I would expect basically correct outputs. It is mildly slower, but not enough for me to notice in the time I did this check. It will just run over extra hits that form tracks and notice they don't look blip like, so it will end up dropping the cluster. 

For case 5:
```
%MSG-s ArtException:  PostEndJob 29-Jan-2026 09:54:03 CST ModuleEndJob
---- EventProcessorFailure BEGIN
  EventProcessor: an exception occurred during current event processing
  ---- ScheduleExecutionFailure BEGIN
    Path: ProcessingStopped.
    ---- LogicError BEGIN
      Invalid FindMany
      ---- ProductNotFound BEGIN
        Found zero products matching all selection criteria
          C++ type: art::Assns<recob::Hit,simb::MCParticle,anab::BackTrackerHitMatchingData>
          Module label: 'blipgaushitTruthMatchwjejefiej'
          Product instance name: ''
          Process name: (empty)
      ---- ProductNotFound END
      Attempt to use a FindMany where the underlying art::Assns product was not found.The above exception was thrown while processing module BlipRecoProducer/blipreco run: 860 subRun: 71 event: 1
    ---- LogicError END
    Exception going through path reco2
  ---- ScheduleExecutionFailure END
---- EventProcessorFailure END
---- FatalRootError BEGIN
  Fatal Root Error: TTree::SetEntries
  Tree branches have different numbers of entries, eg EventAuxiliary has 0 entries while sim::AuxDetSimChannels_genericcrt__G4. has 22 entries.
  ROOT severity: 2000
---- FatalRootError END
%MSG


```

henrylay97 (approved):
Brilliant. Thanks for the detailed checks Jacob! There is always some need for analysers / devs / release management to be careful with fcl labels so the responses to 2 and 3 are reasonable. Overall I'm happy that this will fail gracefully if at all.

As long as you're happy with the production fcl state then I'm happy with this now :D

nathanielerowe:
trigger build ci_ref=v10_14_02_01 LArSoft/lar*@LARSOFT_SUITE_v10_14_02_01 SBNSoftware/sbndaq-artdaq-core@v1_10_06 SBNSoftware/sbn*@SBN_SUITE_v10_14_02_01

nathanielerowe:
Approved

PR #903, opened 2026-01-30
build with geant4 v4_11_2_p02

lgarren:


nathanielerowe:
Can you ping when the version of larsoft that will use geant4 v4_11_2_p02 has been released?

PR #904, opened 2026-02-02
Generic filter module for exclusive cross section sample production

tbwester:
## Description 

Filter module that is configurable for specifying common requirements for cross section signal definitions. This PR is a draft since other options may need to be added.

Features:
- Filter based on final state primary particles with optional KE thresholds
- Filter based on neutrino interaction properties: Nu flavor, In TPC, CC/NC, target PDG codes, modes
- Module checks a list of filters to accept events that pass one or more conditions (e.g., accept events containing particle A or particle B)

Some example signal definition configurations are added to `fcls/signal_filters_sbnd.fcl`.

## Checklist
- [ ] Linked any relevant issues under `Developement`
- [ ] Does this PR affect CAF data format? If so, please assign a CAF maintainer ([PetrilloAtWork](https://github.com/PetrilloAtWork) or [JosiePaton](https://github.com/JosiePaton)) as additional reviewer.
- [ ] Is this PR a patch for the ongoing production? If so, separate PR must also be made for production/v10_06_00 branch!

### DIFF
--- a/sbndcode/Filters/CMakeLists.txt
+++ b/sbndcode/Filters/CMakeLists.txt
@@ -37,6 +37,7 @@ cet_build_plugin(LArG4CRTFilter art::module SOURCE LArG4CRTFilter_module.cc LIBR
 cet_build_plugin(LArG4FakeTriggerFilter art::module SOURCE LArG4FakeTriggerFilter_module.cc LIBRARIES ${MODULE_LIBRARIES})
 cet_build_plugin(SimEnergyDepFakeTriggerFilter art::module SOURCE SimEnergyDepFakeTriggerFilter_module.cc LIBRARIES ${MODULE_LIBRARIES})
 cet_build_plugin(NumberOfHitsFilter art::module SOURCE NumberOfHitsFilter_module.cc LIBRARIES ${MODULE_LIBRARIES})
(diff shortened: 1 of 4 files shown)


afropapp13 (approved):
Many thanks for all the work, happy with the filter!

absolution1:
Hi @tbwester 

This filter looks incredibly useful, and its usefulness extends beyond SBND.

I think this should live in LArSoft, so that other experiments can make use of it (I can already think of some usecases).

The filter could live alongside a LArG4 filter I wrote years ago: https://github.com/LArSoft/larevt/blob/52fa202b71b6186ccc54522651107e9ea7a5724e/larevt/Filters/LArG4ParticleFilter_module.cc

tbwester:
Hi @absolution1 thanks! I considered putting it upstream but then I'm not sure how to implement experiment-specific definitions for the fiducial volume option. Do you know of any way?

absolution1:
hi @tbwester 

Sorry I completely missed your response.  

I had a look at the code, and it looks like its using my old RecoUtils library, I didn't expect this to still be in use.
That function doesn't do anything fancy, it purely ask the geometry service whether its inside of a TPC volume, within a buffer region.

You could copy/paste the function into your filter module.

Is this what you mean?

PR #941, opened 2026-07-10
Feature/blip energy drift correction

Jjm321814:
Adding additional blip energy variables so analyzers have access to a drift corrected energy, as well as an explicitly not drift corrected one.

All drift corrections are anchored to the event timestamp, so the beam blips will come out with the right energy. Any out-of-time blip activity will be mis-corrected, so the noDriftCorrection variables are saved as new variables.


Jjm321814:
This PR is part of a set of 4 in sbndcode, sbnobj, sbncode, and sbnanaobj to make this new blip data entry, and share it to CAF level. 
Each PR is at 
sbndcode: https://github.com/SBNSoftware/sbndcode/pull/941
sbnobj: https://github.com/SBNSoftware/sbnobj/pull/174
sbncode: https://github.com/SBNSoftware/sbncode/pull/661
sbnanaobj: https://github.com/SBNSoftware/sbnanaobj/pull/193

Jjm321814:
Initial checks on MC looked good. The values of energy were very mildly shifted with temporary debug print outs like 
```
 dep El before corr 1767.8
0.0615994
 dep El after corr 1767.88
Blip Time1.55457 Det Lifetime 35000
0.0615994
 finally blip energy 0.0616021
```
Unfortunately the detector properties service used to get the electron lifetime initially applies a fixed lifetime. We will need to connect to some kind of calibration database service to get better results on data.

Jjm321814:
This PR is more complicated. 
Applying the blip drift correction in data required some interface to the NormalizeDriftSQLite_tool. That was added in to the blipAlg files. 
But I don't know how to use an art_tool, so I changed the implementation of NormalizeDriftSQLite_tool so it has an additional class implementation that can be imported more easily. 
To avoid code duplication I split the actual definition and implementation code into their own files that get imported. CMakeLists needed updated to accommodate these changes.

This PR does not interrupt PANDORA's use of the NormalizeDriftSQLite_tool and allows blips to use NormalizeDriftSQLite_class. If we have an example of implementation of the art_tool version (that doesn't result in the code duplication seen in https://github.com/SBNSoftware/sbndcode/blob/27d88f5d147065ed28bd678adfb5ca7e4dd9702e/sbndcode/Calorimetry/LightCaloProducer_module.cc#L821-L848 ) I'd be happy to simplify this PR to that suggestion.

Jjm321814:
Oh oops let me get rid of the print statements I used to check the implementation.

Jjm321814:
I have verified this code imports the correct lifetime values and adds additional variables to the blip larsoft output (reco2) and SRBlip output (CAF). 
The validation was done on 10 data events from one run and 10 MC events.

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
```diff
@@ -268,11 +269,12 @@ namespace blip {
     fCylinderRadius     = pset.get<float>         ("CylinderRadius",      15);
     
     fCaloAlg            = new calo::CalorimetryAlg( pset.get<fhicl::ParameterSet>("CaloAlg") );
+    ElifetimeTool       = new sbnd::calo::NormalizeDriftSQLite( pset.get<fhicl::ParameterSet>("NormalizeDrift"));
```
PetrilloAtWork, line 272: I _recommend_ the use of `std::unique_ptr` here ([`[CF-051]`](https://sbnsoftware.github.io/sbn/codingconv/CodingConventionsExplained.html#CF051)). In shared SBN code, we forbid the use of `new`/`delete`, but this is not SBN code so it's up to you (I would change also the calo algorithm, rather than abiding to its questionable example).
Actually, I would go directly with an object in the class: I currently don't see a reason to allocate it dynamically; but that has also to do with a plan of future extension to other implementations.

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
```diff
@@ -268,11 +269,12 @@ namespace blip {
...
     fCaloPlane          = pset.get<int>           ("CaloPlane",           2);
     fCalodEdx           = pset.get<float>         ("CalodEdx",            2.8);
     fESTAR_p0           = pset.get<float>         ("ESTAR_p0",            0.01730);
     fESTAR_p1           = pset.get<float>         ("ESTAR_p1",            0.00003479);
-    fLifetimeCorr       = pset.get<bool>          ("LifetimeCorrection",  false);
+    fLifetimeCorr       = pset.get<bool>          ("LifetimeCorrection",  true);
```
PetrilloAtWork, line 277: Make sure the change of default is well advertised (also in release notes, if SBND maintains any).

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
```diff
@@ -1141,13 +1144,26 @@ namespace blip {
       // ================================================================================
       float depEl   = std::max(0.0,(double)blip.Charge);
       float Efield  = kNominalEfield;
-
+      float recomb  = ModBoxRecomb(fCalodEdx,Efield);
+      blip.EnergyNoDriftCorrection   = depEl * (1./recomb) * kWion;
```
PetrilloAtWork, line 1148: Why not just
```suggestion
      blip.EnergyNoDriftCorrection   = depEl / recomb * kWion;
```
?
No big deal, but it feels strange.

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
```diff
@@ -1141,13 +1144,26 @@ namespace blip {
...
+      // METHOD 2: recombination factor using dE/dx from NIST tables (dE/dx = kinetic energy / CSDA)
+      float energy_estar = Q_to_E_ESTAR(depEl);
+      float energy_pstar = Q_to_E_PSTAR(depEl);
+      blip.EnergyESTARNoDriftCorrection  = energy_estar;
+      blip.EnergyPSTARNoDriftCorrection  = energy_pstar;
```
PetrilloAtWork, line 1153: It does not look like you really need that temporary parking spot any more:

```suggestion
      blip.EnergyESTARNoDriftCorrection  = Q_to_E_ESTAR(depEl);
      blip.EnergyPSTARNoDriftCorrection  = Q_to_E_PSTAR(depEl)r;
```

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
```diff
@@ -1178,12 +1194,10 @@ namespace blip {
...
-      float recomb  = ModBoxRecomb(fCalodEdx,Efield);
       blip.Energy   = depEl * (1./recomb) * kWion;
-      
       // METHOD 2: recombination factor using dE/dx from NIST tables (dE/dx = kinetic energy / CSDA)
-      float energy_estar = Q_to_E_ESTAR(depEl);
-      float energy_pstar = Q_to_E_PSTAR(depEl);
+      energy_estar = Q_to_E_ESTAR(depEl);
+      energy_pstar = Q_to_E_PSTAR(depEl); //reaculate after drift correction
       blip.EnergyESTAR = energy_estar;
       blip.EnergyPSTAR = energy_pstar;
```
PetrilloAtWork, line 1202:
```suggestion
      blip.EnergyESTAR = Q_to_E_ESTAR(depEl);
      blip.EnergyPSTAR = Q_to_E_PSTAR(depEl); //recalculate after drift correction
```

Jjm321814:
I need to include normalizeDrift settings in the blipconfig.fcl (never pushed those changes)

sbndcode/Calibration/TPCCalorimetry/NormalizeDriftSQLite_Definitions.h
```diff
@@ -0,0 +1,67 @@
...
+// Services
+#include "lardata/DetectorInfoServices/DetectorClocksService.h"
+
+// Lab helpers
+//#include "wda.h"
+
+// C++
+#include <string>
+#include <optional>
+#include <cassert>
```
PetrilloAtWork, line 27: These headers should be sorted out... many of these are not used in the header, so they should live in the `.cc` file.
```suggestion
#include "art/Framework/Principal/Event.h"
#include "fhiclcpp/ParameterSet.h"

#include "larevt/CalibrationDBI/Providers/DBFolder.h"

// Tool include
#include "larreco/Calorimetry/INormalizeCharge.h"

// Services
#include "lardata/DetectorInfoServices/DetectorClocksService.h"

// C++
#include <string>
#include <optional>
#include <cassert>
```

`#include "art/Utilities/ToolMacros.h"` should be moved to the `_tool.h` header (which is where the tool macros are used).
Headers are needed for `recob::Hit` (`lardataobj/RecoBase/Hit.h`), `geo::Point_t` (`larcoreobj/SimpleTypesAndConstants/geo_vectors.h`), `std::map` (`map`) and `uint32_t` (`cstdint`?).

sbndcode/Calibration/TPCCalorimetry/CMakeLists.txt
```diff
@@ -19,7 +19,9 @@ set(TOOL_LIBRARIES
 
 cet_build_plugin(NormalizeDriftSQLite art::tool LIBRARIES ${TOOL_LIBRARIES})
 cet_build_plugin(NormalizeYZ art::tool LIBRARIES ${TOOL_LIBRARIES})
-
+cet_make_library(LIBRARY_NAME sbndcode_GIMME_LIFETIMES
```
PetrilloAtWork, line 22: While `GIMME_LIFETIMES` is indeed pretty straightforward and self-describing, it may complicate users' lifetime. At a certain point, from other code or from the documentation, I find out that I need the class `sbnd::calo::NormalizeDriftSQLite`, and I need to add its library. Normally what I would do is to look for `sbndcode::NormalizeDriftSQLite` or `sbndcode::‎Calibration/TPCCalorimetry` or `sbndcode::‎Calibration/TPCCalorimetry/NormalizeDriftSQLite`, but I will never guess `GIMME_LIFETIMES`.

sbndcode/Calibration/TPCCalorimetry/NormalizeDriftSQLite_class.cc
PetrilloAtWork, line 1: I am confused by this pattern: what is the reason for a `_Definitions.h` and a `_class.h` to coexist? what about `_Implementation.cc` and `_class.cc`?

ups/product_deps
```diff
@@ -253,7 +253,7 @@ wpdir   product_dir     wire-cell-cfg
 #
 ####################################
 product		version		qual	flags		<table_format=2>
-sbncode		v10_21_00       -
+sbncode		v10_21_02       -
```
PetrilloAtWork, line 256: Usually version bumps are not added to the PR, especially for merge in `develop`.
The reason: imagine that now a picky, annoying reviewer (ah-rum) starts giving you hard time for petty reasons. It takes one month to convince him to move forward. At that point, `develop` is using `sbncode` `v10_23_00`: this PR will create a conflict and will need to be fixed. Better to leave this type of updates to the release managers if needed (usually they are not).
```suggestion
sbncode		v10_21_00       -
```

PetrilloAtWork (commented):
Left a few comments. I am particularly confused by the pattern of the tool source code.
However, I bear no authority on `sbndcode`, so mine should be taken just as suggestions.

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
Jjm321814: Updated both

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
Jjm321814: I've let most of the blip analyzers I am aware of know, and most of them prefer this default correction. 
If we are most interested in beam-related activity this gives a better energy value. Additionally the electron lifetime in SBND is so high that the actual changes in energy are very small.

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
Jjm321814: Thats a good point. This exact syntax predates me, but happy to simplify the line.

sbndcode/BlipRecoSBND/Alg/BlipRecoAlg.cc
Jjm321814: Removed both instances of this parking

sbndcode/Calibration/TPCCalorimetry/NormalizeDriftSQLite_Definitions.h
Jjm321814: Moved several to the cc file. I'll verify everything runs fine still.

sbndcode/Calibration/TPCCalorimetry/CMakeLists.txt
Jjm321814: Whoops. I was having trouble getting the cmake sorted and gave myself a really obvious name for printing. I'll update to sbndcode::‎Calibration/TPCCalorimetry/NormalizeDriftSQLite

sbndcode/Calibration/TPCCalorimetry/NormalizeDriftSQLite_class.cc
Jjm321814: The file scheme was to avoid actual code duplication. In particular the DEFINE_ART_CLASS_TOOL call in _tool.h causes issues for treating the code as a standalone class. 
_implementation.cc has the actual code that is shared between the tool and class instance. 

Since the _implementation.h is shared by _tool.h and _class.h it can't have ifndef protections, so we need both class.h and tool.h. 

So this splitting is primarily that I don't understand how to use an Art_Tool, and this allows both a class and a tool implementation to exist without any code duplication.

ups/product_deps
Jjm321814: We do have a conflict to sort now. My test are was in v10_21_00 and I think I added this with a "git add -u". Is it better to specifically exclude this file from future PR?

Jjm321814:
Responded to the comments. I'll test the area again, and then need to pull the changes into the production branch PR

Jjm321814:
Code output looks good after above fixes. I will get these rolled into the related production PR

Jjm321814:
All four of these related PR should be ready to merge now.

PR #961, opened 2026-09-22
Add QE neutral hyperon filter and lambda charged decay filter

SergioYB:
## Description 
I have included two filters:
- The first one is a GENIE filter based on the `TrueSignalFilter` module adding the `sbnd_qe_hyperon_filter` block to `signal_filters_sbnd.fcl` and the corresponding modification on `run_sbnd_gen2_rare_signal_filter.fcl`.
- The second one is a new module, `DecayFilterG4`, that is run after Geant4 and checks if a given mother particle decays into a specific set of daughter particles. In my case it is used to filter the charged decay channel of Lambda hyperons and uses the configuration of the `sbnd_lambda_charged_decay_filter` block found in `g4_decay_filters.fcl`. This filter is run using `run_sbnd_g4_lambda_decay_filter.fcl`.

I have previously used both filters to simulate the QE Lambda signal, and I confirm that, after running the filters, the expected target events are present in the output. I have double-checked by running the filters again on some of my  QE Lambda samples and all events pass them.

$${\color{blue}!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!}$$

$${\color{blue}\bf{\textrm{IMPORTANT UPDATE Feb 2nd 2026:}}}$$  If you are making a PR which is intended as a patch for the CURRENT production for gen 2 SBND analyses, you must make two PRs: one for develop and one for the production/sbnd-gen2 branch.

$${\color{blue}!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!}$$

## Checklist
- [ ] Linked any relevant issues under `Developement`
- [ ] Does this PR affect CAF data format? If so, please assign a CAF maintainer ([PetrilloAtWork](https://github.com/PetrilloAtWork) or [tjones79](https://github.com/tjones79)) as additional reviewer.
- [ ] Is this PR a patch for the ongoing production? If so, separate PR must also be made for production/v10_06_00 branch! 

No

N/A

**GENIE filter performance metrics:**
- **Filter efficiency:** ~-8.1e-3% -~ 0.0079% (forgot to remove hyperons from anti-nue interactions)
- **Average CPU Time (hyperon filter module):** ~0.032 s/event

**G4 filter performance metrics:**
- **Filter efficiency (after applying GENIE filter):** ~62% 
- **Average CPU Time (filter module):** ~0.078 s/event

*NOTES*
- Due to the low efficiency of the GENIE filter, and in order to avoid simulating a large number of events just to evaluate the filter, the efficiency was estimated using Aurora dev samples by measuring the fraction of events fulfilling the filter requirements.
- To estimate CPU time I have run the filter on 490 Reco1 events from gen2 dev Sample


tbwester:
Hi @SergioYB, thanks for adding this feature! Two comments, both along similar lines: I think since there aren't that many hyperons we could get away with saving all of them, so I think we don't need a separate QE-only filter, even if that's what the analysis is looking for (besides, non-QE may be a background). And similarly for decay modes, can we just save all the decay modes, instead of only particular ones? That would simplify the workflow and make the sample more generally useful.

SergioYB:
Hi @tbwester, thanks for the comments. Regarding the first point, the QE hyperon channel is quiet suppressed (~12% of all hyperon production), while lambdas are produced in ~50% of the QE cases and the charged decay occurs ~62% of the time. Since my current selection efficiency is ~4% and I would like to study some extra cuts to improve the purity, I would prefer to have an enhanced QE neutral hyperon sample. I am focusing on neutral hyperons because sigma0 could be a potential background, whereas sigma- can be easily removed applying gap-based cuts.

The remaining hyperon production channels have a higher particle multiplicity, and the cuts seem to effectively remove them. For example, none of the ~7k non-QE hyperon produced inside the TPC in the aurora dev sample pass the selection. In addition, Thomas Jones told me that he has also requested a general hyperon sample, which I could use to increase these backgrounds. 

Regarding the decay filter, my selection is based on the topology of the charged channel (proton + pi-). The other lambda decay channel is into a neutron and pi0, whose signature is very different from the charged one, so it is very unlikely to pass the selection cuts. However, I don't mind if this second cut is not applied. I was only suggesting it to obtain a purer sample, but it is not essential for my analysis.

